PR #84, opened 2016-10-13
MULE-10274: Move extension model to the runtime so it can be used to …

marianogonzalez:
…describe any extension.

### DIFF
--- a/src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ExtensionDeclarer.java
+++ b/src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ExtensionDeclarer.java
@@ -1,0 +1,260 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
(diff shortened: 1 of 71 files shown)


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ExtensionDeclarer.java
```diff
@@ -0,0 +1,260 @@
...
+  private boolean isTypeRegistrable(ObjectType objectType) {
+    String typeId = getTypeId(objectType).orElse(null);
+    if (typeId != null &&
+        (Object.class.getName().equals(typeId) ||
+            typeId.startsWith("org.mule.runtime.") ||
+            typeId.startsWith("com.mulesoft.mule.runtime."))) {
+      return false;
+    }
+
+    return true;
```
alegmarra: how can a type without TypeId be registerable?

marianogonzalez: Just like it is today. I think this is related to the fact that just like we did with most model properties, most type annotations will be moved to the MetadataType , so I wouldn't worry too much about this

PR #86, opened 2016-10-17
MULE-10104: Start development of the new WS consumer extension

juandesi:


marianogonzalez:
It doesn't make sense to commit into this project with that issue description. Use an existing (or new) jira issue which describes what you're doing

marianogonzalez:
same comment as with the mule-api one

PR #88, opened 2016-10-18
MULE-10543: Implement the SchedulerService

elrodro83:


src/main/java/org/mule/runtime/api/scheduler/SchedulerService.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Set;
+
+/**
+ * Provides access to the different schedulers and thread pools that exist in the Mule runtime, allowing an artifact to schedule
+ * tasks on those.
+ * <p>
+ * This interface provides access to different schedulers, each with its own tuning options, optimized for some specific kind of
+ * work. Artifacts that need to dispatch tasks need to carefully determine what's the scheduler that best fits the nature of the
+ * task in order to keep the resources usage and the impact on other artifacts at a minimum.
+ *
```
dfeist: Where are their defaults configured?

src/main/java/org/mule/runtime/api/scheduler/SchedulerService.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Builds a fresh {@link Scheduler} for light CPU tasks. The returned {@link Scheduler} is backed by the Mule runtime cpu-light
+   * executor.
+   * <p>
+   * A task is considered {@code cpu-light} if it doesn't block at any time and its duration is less than 10 milliseconds.
+   * 
+   * @return a scheduler for cpu-light tasks
+   */
+  Scheduler cpuLightScheduler();
```
dfeist: Let's update this naming later if we think of something better. Also let's make sure it matches whatever we have in ProcessinType enum. ( i think it does currently)

src/main/java/org/mule/runtime/api/scheduler/SchedulerService.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Builds a fresh {@link Scheduler} for blocking I/O tasks. The returned {@link Scheduler} is backed by the Mule runtime
+   * blocking I/O executor.
+   * <p>
+   * A task is considered {@code I/O} if it spends most of it's clock duration blocked due to I/O operations.
+   * 
+   * @return a scheduler for I/O tasks
+   */
+  Scheduler ioScheduler();
```
dfeist: Why do these return new instances, I thought the idea was that the scheduler service returning shared instances primarily, and had some seperate methods for creating new pools.

src/main/java/org/mule/runtime/api/scheduler/SchedulerService.java
```diff
@@ -0,0 +1,62 @@
...
+   * @return a scheduler for computation tasks
+   */
+  Scheduler computationScheduler();
+
+  /**
+   * Provides reporting and statistical information about the {@link Scheduler}s maintained by this service instance.
+   * 
+   * @return a set where each element contains the information of an individual {@link Scheduler}.
+   */
+  Set<SchedulerStats> getSchedulerStats();
```
dfeist: Does it make sense to separate this rather than have schedulers expose this? If it is separate doesn't a map make more sense than a set, or how are schedulers identified?

src/main/java/org/mule/runtime/api/scheduler/Scheduler.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * Provides access to a specific executor in the Mule runtime.
+ * <p>
+ * See {@link ScheduledExecutorService} and {@link ExecutorService} for documentation on the provided methods.
+ * 
+ * @since 1.0
+ */
+public interface Scheduler extends ScheduledExecutorService {
```
dfeist: Is `ScheduledExecutorService` exactly what we need, or it's just convenient?  Do we also need some kind of identifier?

dfeist: I'm not convinced all schedulers should support scheduling work, but i can't remember what we discussed about this either.

elrodro83: The consensus was that we needed it.

src/main/java/org/mule/runtime/api/scheduler/SchedulerService.java
elrodro83: Will update the javadocs

PR #93, opened 2016-11-01
MULE-10349: Support expression for "response" parameters in sources

marianogonzalez:


src/main/java/org/mule/runtime/api/metadata/MediaType.java
```diff
@@ -9,7 +9,7 @@
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.list;
 import static java.util.Optional.ofNullable;
-import static org.mule.metadata.internal.utils.StringUtils.isNotEmpty;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
```
pablolagreca: I thought we were not depending on anything else than the metadata-api. Can you check why are you being able to use StringUtils?

src/main/java/org/mule/runtime/api/metadata/MetadataKeyBuilder.java
```diff
@@ -7,7 +7,7 @@
 package org.mule.runtime.api.metadata;
 
 import static java.util.stream.Collectors.toSet;
-import static org.mule.metadata.internal.utils.StringUtils.isNotEmpty;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
```
pablolagreca: same here

PR #100, opened 2016-11-14
MULE-10885: Support operation UX

marianogonzalez:


src/main/java/org/mule/runtime/api/meta/model/parameter/ParameterRole.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @since 1.0
+ */
+public enum ParameterRole {
+
+  /**
+   * Indicates that the parameter is only configuring the operation and doesn't represent
+   * data that is being sent to another endpoint
+   */
+  PARAMETERIZATION,
```
pablolagreca: Any other idea?. Perhaps CONFIGURATION ? All parameters are parameterization.

marianogonzalez: i discarded that option because it seemed to me like it could be confusing with a parameter which actually is the extension's configuration

PR #101, opened 2016-11-14
MULE-10943: Parameter placement order have two different default values

marianogonzalez:


src/main/java/org/mule/runtime/api/meta/model/display/LayoutModel.java
```diff
@@ -131,7 +133,7 @@ public static LayoutModelBuilder builderFrom(LayoutModel template) {
   private boolean password = false;
   private boolean text = false;
   private boolean query = false;
-  private int order;
+  private int order = DEFAULT_ORDER;
```
alegmarra: having a "big number" here doesn't mean anything for the tooling. I think that `-1` was better in the sense that we could communicate to the UI that no specific order was declared, so the y could do whatever they wanted with that field's order

PR #103, opened 2016-11-15
MULE-10943: Parameter placement order have two different default values

marianogonzalez:


src/main/java/org/mule/runtime/api/meta/model/display/LayoutModel.java
```diff
@@ -19,6 +24,8 @@
  */
 public final class LayoutModel {
 
+  public static final int DEFAULT_ORDER = -1;
```
estebanwasinger: Marra wins

PR #104, opened 2016-11-17
MULE-10977: java.lang.Object is persisted as a type in the extension …

pablomusumeci:
…model JSON


marianogonzalez:
Rejected on the same bases as the related Metadata one

PR #110, opened 2016-11-25
MULE-10861: Support new Error Handling in Extensions API

estebanwasinger:


src/main/java/org/mule/runtime/api/error/Errors.java
```diff
@@ -0,0 +1,67 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.error;
+
+/**
+ * Utility class to list all the commons error types that Mule and the components inside of it
+ * could throw.
+ *
+ * @since 4.0
```
pablomusumeci: Is this 4.0 or 1.0?

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/OperationDeclaration.java
```diff
@@ -16,11 +20,20 @@
...
+
   /**
    * {@inheritDoc}
    */
   OperationDeclaration(String name) {
     super(name);
   }
 
+  public void addErrorType(ErrorModel errorModel) {
+    errorModels.add(errorModel);
```
marianogonzalez: This is wrong. You're requiring an ErrorModel but don't provide a way to create it. You should follow either of two approaches:

* You create an ErrorDeclarer (with its backing ErrorDeclaration) and it is the ExtensionModelFactory the one who knows how to instantiate the concrete ErrorModel
* You take the approach of the LayoutModel or the DisplayModel and you give such model a builder.

I think that option number with is the one that fits this case better

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/OperationDeclaration.java
```diff
@@ -16,11 +20,20 @@
...
+  private Set<ErrorModel> errorModels = new HashSet<>();
+
   /**
    * {@inheritDoc}
    */
   OperationDeclaration(String name) {
     super(name);
   }
 
+  public void addErrorType(ErrorModel errorModel) {
```
marianogonzalez: the method name is not good. You add a type but then receive a model.... this is weird

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/OperationDeclarer.java
```diff
@@ -22,4 +24,12 @@
     super(declaration);
   }
 
+  /**
+   * Adds an {@link ErrorModel} to indicate that the current operation could throw the added error.
+   *
+   * @param error {@link ErrorModel} to add to the {@link OperationDeclaration}
+   */
+  public void withErrorType(ErrorModel error) {
```
marianogonzalez: withErrorType -> withError

src/main/java/org/mule/runtime/api/meta/model/error/ErrorModel.java
```diff
@@ -0,0 +1,49 @@
...
+import org.mule.runtime.api.error.Errors;
+import org.mule.runtime.api.meta.model.operation.OperationModel;
+
+import java.util.Optional;
+
+/**
+ * A model which represents a error possible error that could be thrown
+ * by the Operation represented by the {@link OperationModel}.
+ * </p>
+ * This model declares:
```
marianogonzalez: do not repeat what the other methods already say. Otherwise, you have to change this jdoc each time you add something

src/main/java/org/mule/runtime/api/meta/model/error/ThrowsErrors.java
```diff
@@ -0,0 +1,24 @@
...
+ * @since 1.0
+ */
+public interface ThrowsErrors {
+
+  /**
+   * @return a {@link Set} of {@link ErrorModel} with the possible errors that the current
+   * component could throw.
+   * @see ErrorModel
+   */
+  Set<ErrorModel> getErrorTypes();
```
marianogonzalez: OperationModel and  ConnectionProviderModel should implement this interface

src/main/java/org/mule/runtime/api/meta/model/error/ErrorModel.java
```diff
@@ -0,0 +1,49 @@
...
+package org.mule.runtime.api.meta.model.error;
+
+import org.mule.runtime.api.error.Errors;
+import org.mule.runtime.api.meta.model.operation.OperationModel;
+
+import java.util.Optional;
+
+/**
+ * A model which represents a error possible error that could be thrown
+ * by the Operation represented by the {@link OperationModel}.
```
alegmarra: if this is for sources too, then it's ComponentModel

src/main/java/org/mule/runtime/api/meta/model/error/ErrorModel.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  String getType();
+
+  /**
+   * Gets the namespace of error.
+   * This namespace represent the origin or how declares this error.
+   *
+   * @return The namespace of the error
+   */
+  String getNamespace();
```
alegmarra: what's the use case for an imported error? is that kind of complexity really required?

alegmarra: I think we should enforce the fact that each extension provides it's own custom errors

estebanwasinger: Imported error types are not supported, extension's errors can only inherit from `Mule Errors`

PR #113, opened 2016-11-30
MULE-11105: Move Scheduler interface to mule-api

elrodro83:


src/main/java/org/mule/runtime/api/scheduler/Scheduler.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Allows tasks to be submitted/scheduled to a specific executor in the Mule runtime. Different {@link Scheduler} instances may be
+ * backed by the same {@link ExecutorService}, allowing for a fine control of the source of the tasks that the underlying
+ * {@link ExecutorService} will run.
+ * <p>
+ * See {@link ScheduledExecutorService} and {@link ExecutorService} for documentation on the provided methods.
+ * 
+ * @since 4.0
```
marianogonzalez: 1.0

elrodro83:
@marianogonzalez done

src/main/java/org/mule/runtime/api/scheduler/Scheduler.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Tries to do a graceful shutdown.
+   * <p>
+   * If this hasn't terminated after a configured time, a forceful shutdown takes place.
+   * 
+   * @param gracefulShutdownTimeout the maximum time to wait for the running tasks to gracefully complete.
+   * @param unit the time unit of the {@code timeout} argument
+   */
+  void stop(long gracefulShutdownTimeout, TimeUnit unit);
```
dfeist: IMO we need to expose a stop() without parameters here, and have the service manage the timeout.

src/main/java/org/mule/runtime/api/scheduler/Scheduler.java
```diff
@@ -0,0 +1,83 @@
...
+   *         upon cancellation
+   * @throws RejectedExecutionException if the task cannot be scheduled for execution
+   * @throws NullPointerException if command is null
+   */
+  public ScheduledFuture<?> scheduleWithCronExpression(Runnable command, String cronExpression, TimeZone timeZone);
+
+  /**
+   * @return The {@link ThreadType} that matches with the {@link Thread}s managed by this {@link Scheduler}.
+   */
+  ThreadType getThreadType();
```
dfeist: I'd take this out, we have no real use for it currently in API, so best if we leave it out for now.

src/main/java/org/mule/runtime/api/scheduler/Scheduler.java
elrodro83: The thing is that te timeout comes from the configuration of the application, accessible from the muleContext. Making this change would be part of a larger story, where that configuration is moved from the application to the runtime.

PR #115, opened 2016-12-07
MULE-10760: HTTP Service - Create request and response builders

afelisatti:
- Move case insensitive map solution to api

### DIFF
--- a/src/main/java/org/mule/runtime/api/util/CaseInsensitiveMapWrapper.java
+++ b/src/main/java/org/mule/runtime/api/util/CaseInsensitiveMapWrapper.java
@@ -1,0 +1,258 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/util/CaseInsensitiveMapWrapper.java
```diff
@@ -0,0 +1,258 @@
...
+ * <p/>
+ * When a key/value pair is put in the map the key case is remembered so when the key set or the entry set is retrieved the
+ * correct case is returned. This is useful to store, for example, camel case keys. However, two keys that only differ in their
+ * case will be assumed to be the same key and only one value (the last) will be kept. Note: as this map uses a provided class to
+ * create the backing map, key rewrite is not ensured. It is possible that when redefining a value associated to a key, the key
+ * case won't be overwritten and the already existing key case will remains in the key set and entry set.
+ *
+ * @param <T> The class of the values referenced in the map.
+ *
+ * @since 3.6.0
```
gsfernandes: Shouldn't this be changed to 1.0?

PR #120, opened 2016-12-21
MULE-11214: Add concept of "resolverName" to the NamedTypeResolver

pablomusumeci:


src/main/java/org/mule/runtime/api/metadata/resolving/AttributesTypeResolver.java
```diff
@@ -32,4 +32,11 @@
    * @throws ConnectionException if an error occurs when using the connection provided by the {@link MetadataContext}
    */
   MetadataType getAttributesType(MetadataContext context, K key) throws MetadataResolvingException, ConnectionException;
+
+  /**
+   * {@inheritDoc}
+   */
+  default String getResolverName() {
+    return "attributes";
```
marianogonzalez: @juandesi shouldn't this be the enum you defined?

juandesi: yes, rebase from master and you will have a new MetadataComponent enum to use in this default values.

PR #122, opened 2016-12-23
MULE-10413: Define how to define EL functions from modules

afelisatti:


src/main/java/org/mule/runtime/api/metadata/DataTypeBuilder.java
```diff
@@ -107,4 +134,25 @@
 
   }
 
+  /**
+   * Provides methods to set data associated to the resources of an {@link ExpressionFunction}.
+   */
+  interface DataTypeFunctionTypeBuilder extends DataTypeParamsBuilder {
+
+    /**
+     *
```
gsfernandes: Add description to the method.

PR #127, opened 2016-12-31
MULE-11237: Add language rules into the extension model

marianogonzalez:


src/main/java/org/mule/runtime/api/meta/model/parameter/ParameterGroupModel.java
```diff
@@ -33,6 +33,11 @@
   String ADVANCED = "Advanced";
 
   /**
+   * Group name for parameters that are for controling the opertion's output
+   */
+  String OUTPUT = "Advanced";
```
alegmarra: Advanced => Output

marianogonzalez: fixed

PR #128, opened 2016-12-31
MULE-11238: Explicit parameter groups in DSL

alegmarra:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterGroupDeclaration.java
```diff
@@ -71,4 +72,12 @@ public LayoutModel getLayoutModel() {
   public void setLayoutModel(LayoutModel layoutModel) {
     this.layoutModel = layoutModel;
   }
+
+  public void showInline(boolean showInline) {
```
marianogonzalez: jdoc

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterGroupDeclarer.java
```diff
@@ -52,6 +52,11 @@ public OptionalParameterDeclarer withOptionalParameter(String name) {
     return this;
   }
 
+  public ParameterGroupDeclarer<D> withInlineDefinition(boolean showInline) {
```
marianogonzalez: jdoc and clarify method name

PR #130, opened 2017-01-02
MULE-11320 - Add stop capability to ExtensionWalker

alegmarra:


src/main/java/org/mule/runtime/api/meta/model/util/ExtensionWalker.java
```diff
@@ -42,21 +45,33 @@
...
+    ifContinue(() -> walkOperations(extensionModel));
+  }
+
+  /**
+   * When invoked, the traversal of the {@link ExtensionModel} will
+   * stop once the current visit is completed.
+   * This will not break the execution within the invoking method,
+   * but instead avoid further traversal of the model from that point on.
+   */
+  public final void stop() {
```
marianogonzalez: shouldn't this be protected? Only the implementation logic should be able to stop it

PR #131, opened 2017-01-03
MULE-11325: Change policy descriptor to use json instead of propertie…

pablokraan:
…s files

_ Extracting AbstractMuleArtifactModel from MulePluginModel
_ Extracting model builder from MulePluginModel into AbstractMuleArtifactModelBuilder
_ Extracting AbstractMuleArtifactModelJsonSerializer from MulePolicyModelJsonSerializer
_ Adding model, builder and serializer for policies

### DIFF
--- a/src/main/java/org/mule/runtime/api/deployment/meta/AbstractMuleArtifactModel.java
+++ b/src/main/java/org/mule/runtime/api/deployment/meta/AbstractMuleArtifactModel.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.api.deployment.meta;
+
+import static java.util.Optional.ofNullable;
+
+import org.mule.runtime.api.meta.model.ExtensionModel;
+
+import java.util.Optional;
+
+/**
+ * Base class for creating models for Mule artifacts from json describer files.
+ * <p/>
+ * The idea behind this object is to just load some bits of information that later each "loader" will consume to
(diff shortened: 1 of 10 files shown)


src/main/java/org/mule/runtime/api/deployment/meta/AbstractMuleArtifactModel.java
```diff
@@ -0,0 +1,56 @@
...
+package org.mule.runtime.api.deployment.meta;
+
+import static java.util.Optional.ofNullable;
+
+import org.mule.runtime.api.meta.model.ExtensionModel;
+
+import java.util.Optional;
+
+/**
+ * Base class for creating models for Mule artifacts from json describer files.
```
fernandezlautaro: json > JSON

src/main/java/org/mule/runtime/api/deployment/meta/AbstractMuleArtifactModel.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  protected AbstractMuleArtifactModel(
+                                      String name, String minMuleVersion,
+                                      MuleArtifactLoaderDescriptor classLoaderModelLoaderDescriptor) {
+    this.minMuleVersion = minMuleVersion;
+    this.classLoaderModelLoaderDescriptor = classLoaderModelLoaderDescriptor;
+    this.name = name;
+  }
+
+  public String getName() {
```
fernandezlautaro: missing javadoc 
I believe the only important thing to denote on `name` javadoc is that it's used **only** for logging purposes, (I'm not sure if it's enough with the constructor documentation though)

pablokraan: Name is not used for logging only, is used to construct the artifact descriptor. Besides that, adding javadocs to getters that will only say "returns the artifact model name" seems redundant to me

PR #134, opened 2017-01-06
MULE-11222 - Retrieve ExtensionModel from ComponentIdentifier

alegmarra:


src/main/java/org/mule/runtime/api/dsl/model/ApplicationElement.java
```diff
@@ -0,0 +1,164 @@
...
+   * @return a {@code java.util.List} of all the child {@code ComponentModel}s
+   */
+  public List<ApplicationElement> getInnerComponents() {
+    return innerElements;
+  }
+
+  /**
+   * @return content of the configuration element.
+   */
+  public Optional<String> getValue() {
```
estebanwasinger: This returns the String DSL representation of the element?

src/main/java/org/mule/runtime/api/dsl/model/ApplicationElementIdentifier.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.mule.runtime.api.util.Preconditions.checkState;
+
+/**
+ * Unique identifier for a configuration option. Every configuration option has a namespaceURI and an identifier.
+ *
+ * The namespace defines the extension that declares the component.
+ *
+ * @since 1.0
+ */
+public class ApplicationElementIdentifier {
```
estebanwasinger: Already exist a class with a very similar purpose:  `ComponentIdentifier`, is this useful for you?

marianogonzalez: how is this different from ComponentIdentifier? When should each be used?

src/main/java/org/mule/runtime/api/dsl/model/ApplicationElementIdentifier.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param name unique identifier within the namespace of the language configuration extension
+     * @return the builder
+     */
+    public Builder withName(String name) {
+      identifier.name = name;
+      return this;
+    }
+
+    public ApplicationElementIdentifier build() {
+      checkState(identifier.namespaceUri != null && !identifier.namespaceUri.isEmpty(), "Namespace URI must be not blank");
```
marianogonzalez: use `trim()` in these validations

src/main/java/org/mule/runtime/api/meta/model/ParameterDslConfiguration.java
```diff
@@ -84,10 +84,10 @@ public static ElementDslModelBuilder builder() {
   }
 
   /**
-   * @param prototype a prototype {@link ElementDslModel}
+   * @param prototype a prototype {@link ParameterDslConfiguration}
    * @return a new builder initialised to match the state of the given {@code prototype}
    */
-  public static ElementDslModelBuilder builder(ElementDslModel prototype) {
+  public static ElementDslModelBuilder builder(ParameterDslConfiguration prototype) {
```
marianogonzalez: Sholdn't this builder be renamed to?

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterDeclaration.java
```diff
@@ -32,7 +32,7 @@
   private MetadataType type;
   private boolean hasDynamicType;
   private Object defaultValue = null;
-  private ElementDslModel dslModel = getDefaultInstance();
+  private ParameterDslConfiguration dslModel = getDefaultInstance();
```
marianogonzalez: dslModel -> dslConfiguration

Although in all honesty, the prior name seemed more clear to me

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/util/DeclarationWalker.java
```diff
@@ -132,6 +132,8 @@ private void walkSources(WithSourcesDeclaration declaration) {
       SourceDeclaration sourceDeclaration = (SourceDeclaration) source;
       onSource(declaration, sourceDeclaration);
       ifContinue(() -> walkParameters(sourceDeclaration));
+      ifContinue(() -> sourceDeclaration.getSuccessCallback().ifPresent(this::walkParameters));
```
marianogonzalez: didn't @pablomusumeci  added this same things a couple of days ago?

pablomusumeci: Yes indeed. https://github.com/mulesoft/mule-api/blob/master/src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/util/DeclarationWalker.java#L135

src/main/java/org/mule/runtime/api/dsl/model/ApplicationElementIdentifier.java
alegmarra: ComponentIdentifier has some behaviour related to how we handle the ComponentModels internally in the spring-config module, while this is a little more canonical (the ComponentIdentifier uses mostly prefixes as the namespace instead of using the URI, or defaults to 'mule' if no prefix was provided).
That said, it covers almost identical purposes and the ComponentIdentifier will most likely be removed when [MULE-11355](https://www.mulesoft.org/jira/browse/MULE-11355) is implemented. For now I needed an identifier in the API that matched the ApplicationElement concept.
I'll add a TODO with the jira also in the ComponentIdentifier.

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterDeclaration.java
alegmarra: DslModel has much more meaning now since it relates the DSL representation with the ExtensionModel part of a given node. This provides a way to configure some of the properties that will later be propagated to the DSL, so that's the reason of the change.

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/util/DeclarationWalker.java
alegmarra: yeah, I recall merging this on a rebase, don't now why it's shown here

src/main/java/org/mule/runtime/api/dsl/model/ApplicationElement.java
alegmarra: no, this will provide the value of the dom Element. 
If this ApplicationElement represents an attribute in the xml, then this value will be it's string value, and for the case of nested elements we can have no value (this represents a node that has child elements) or the value of the node's textContent (think a node that represents an @Content paramter)

alegmarra: I'll update the jdocs :)

PR #136, opened 2017-01-09
MULE-11364: Replace TypedValue interface on mule-api with

elrodro83:
DefaultTypedValue from mule-core

### DIFF
--- a/src/main/java/org/mule/runtime/api/metadata/TypedValue.java
+++ b/src/main/java/org/mule/runtime/api/metadata/TypedValue.java
@@ -6,6 +6,8 @@
  */
 package org.mule.runtime.api.metadata;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/metadata/TypedValue.java
```diff
@@ -6,24 +6,50 @@
...
+  /**
+   * Constructs a new {@link TypedValue} with the given parameters.
+   * 
+   * @param value this object's content.
+   * @param dataType the {@link DataType} for this object's content.
+   */
+  public TypedValue(T value, DataType dataType) {
+    this.value = value;
+    if (dataType == null) {
+      this.dataType = DataType.fromObject(value);
```
pablokraan: static import?

PR #138, opened 2017-01-16
MULE-11244: Expose Metadata result information such as resolver and c…

pablomusumeci:
…ategory names in the MetadataService


src/main/java/org/mule/runtime/api/metadata/MetadataAttributes.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.mule.runtime.api.metadata;
+
+import java.util.Optional;
+
+/**
+ * @since 1.0
+ */
+public interface MetadataAttributes {
```
marianogonzalez: jdoc in class and methods

src/main/java/org/mule/runtime/api/metadata/descriptor/ImmutableMetadataAttributes.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+
+public final class ImmutableMetadataAttributes implements MetadataAttributes {
+
+  private final Optional<MetadataKey> key;
+  private final String category;
+  private final Map<String, String> parameters;
+  private final Optional<String> outputResolver;
```
marianogonzalez: don't have Optionals as state. Have the plain field and build the optional on the getter

src/main/java/org/mule/runtime/api/metadata/descriptor/ImmutableMetadataAttributes.java
```diff
@@ -0,0 +1,59 @@
...
+  private final Optional<String> attributesResolver;
+
+  public ImmutableMetadataAttributes(Optional<MetadataKey> key, String category,
```
marianogonzalez: this class should not be API. Move to internal package

src/main/java/org/mule/runtime/api/metadata/descriptor/ImmutableComponentMetadataDescriptor.java
```diff
@@ -6,44 +6,31 @@
...
  * Immutable concrete implementation of {@link ComponentMetadataDescriptor}
  *
  * @since 1.0
  */
-public final class ImmutableComponentMetadataDescriptor implements ComponentMetadataDescriptor {
-
-  private final String name;
-  private final InputMetadataDescriptor input;
-  private final OutputMetadataDescriptor output;
+public final class ImmutableComponentMetadataDescriptor<T extends ComponentModel> implements ComponentMetadataDescriptor<T> {
```
marianogonzalez: this class should not be api. Move to an internal package

src/main/java/org/mule/runtime/api/metadata/descriptor/builder/MetadataDescriptorBuilder.java
```diff
@@ -19,11 +21,18 @@
 public abstract class MetadataDescriptorBuilder {
```
marianogonzalez: This class is no longer a builder but a Builder factory. This class is not making much sense to me anymore. Maybe add a static builder method to each of the components?

pablomusumeci:
I know, but while I'm writing them I wanna you to check the code ;)

src/main/java/org/mule/runtime/api/metadata/descriptor/builder/MetadataAttributesBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.mule.runtime.api.metadata.MetadataAttributes;
+import org.mule.runtime.api.metadata.MetadataKey;
+import org.mule.runtime.api.metadata.descriptor.ImmutableMetadataAttributes;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class MetadataAttributesBuilder {
```
marianogonzalez: What's the criteria here? Why having MetadataAttributes separated in interface and impl but then have the builder stand on its own? Unify interface and impl for all these descriptors or also separate the builders

PR #141, opened 2017-01-18
MULE-11459: Introduce TraversableStream

marianogonzalez:


src/main/java/org/mule/runtime/api/util/func/UnsafeRunnable.java
```diff
@@ -0,0 +1,24 @@
...
+package org.mule.runtime.api.util.func;
+
+/**
+ * Similar to {@link Runnable} but for operations which might throw
+ * a {@link Exception}
+ *
+ * @since 1.0
+ */
+@FunctionalInterface
+public interface UnsafeRunnable {
```
dfeist: A runnable, almost by definition is, fire and forget.  Why not use a `Callable` with void return which in turn is then compatible with `ExecutorService` (I see you wrapping currently) and much easier to integrate with RX libraries?.  The other issue I have with this is that "unsafe" in Java is used typically used to refer to use of `sun.misc.Unsafe`, not the fact something throws an exception.

The other option is RuntimeExceptions of course.

src/main/java/org/mule/runtime/api/stream/bytes/TraversableStream.java
```diff
@@ -0,0 +1,251 @@
...
+ * to reference different positions on the buffer and access them in a thread safe manner.
+ * <p>
+ * This stream is not closed until all open cursors are closed. This makes it critical for anyone invoking the
+ * {@link #openCursor()} method to make sure that the cursor is closed when no longer needed, since otherwise
+ * any resources held by the underlying buffer will never be released.
+ *
+ * @see CursorStream
+ * @since 1.0
+ */
+public class TraversableStream extends InputStream {
```
dfeist: Does nothing like this already exist anywhere?  I assume it doesn't, but it'd be good (for the future) to comment on other implementations that may have been ruled out, and what this does that others don't in javadoc.

src/main/java/org/mule/runtime/internal/stream/bytes/FileStoreTraversableBuffer.java
```diff
@@ -0,0 +1,266 @@
...
+ * <p>
+ * Another possible scenario, is one in which the data requested is behind the buffer's current position, in which case
+ * the data is obtained by reading the temporal file.
+ * <p>
+ * In either case, what's really important to understand is that the buffer is <b>ALWAYS</b> moving forward. The buffer
+ * will never go back and reload data from the temporal file. It only gets data from the stream.
+ *
+ * @since 1.0
+ */
+public final class FileStoreTraversableBuffer extends BaseTraversableBuffer {
```
dfeist: Do these implementation need to be in mule-api?

src/main/java/org/mule/runtime/api/util/func/UnsafeRunnable.java
marianogonzalez: This class already exists in mule, i'm just moving it. To me Runnable is not about fire and forget but about not expecting a return type. There're places where you need a lambda like this and Java limitations to handle checked exceptions become a problem. I don't see a problem with this since you never actually see the class anyway.

src/main/java/org/mule/runtime/api/stream/bytes/TraversableStream.java
marianogonzalez: This is based on the SeekableStream that DW has. However SeekableStream is single threaded only and written Scala. This class is a java migration of that which has been extensively modified to support concurrent access.

JDocs should only explain what the class does. Discussions about ruled out options, etc. belong in a technical document (I think we're even standarizing on that now)

src/main/java/org/mule/runtime/internal/stream/bytes/FileStoreTraversableBuffer.java
marianogonzalez: At least for now yes. DW needs to be able to use it.

src/main/java/org/mule/runtime/api/util/func/UnsafeRunnable.java
dfeist: I didn't realise it was in core, but regardless, that doesn't mean we shouldn't review what we push up into API.  My issues with was that we're using our own interface for something you do with `Callable<Void>` and while  you might avoid `return null` you will are going to need to wrap in Callable/Runnable to use with anything else.

**But,** that said, with lambdas using Callable is a pain for this, and wrapping with a callbale/runnable lambda if/when needed is fairly trivial.  I still don't like the name, but it's thats what it's already called, then so be it.

src/main/java/org/mule/runtime/api/stream/bytes/TraversableStream.java
marianogonzalez: I know, but ConcurrentRAMOffHeapableStream sounded like a bit too much. I also considered SuperAwesomeStream but Kraan wouldn't merge it

src/main/java/org/mule/runtime/internal/stream/bytes/FileStoreTraversableBuffer.java
```diff
@@ -0,0 +1,266 @@
...
+  public FileStoreTraversableBuffer(InputStream stream, byte[] available, int bufferSize) {
+    super(bufferSize);
+    this.stream = stream;
+    this.buffer = allocateDirect(bufferSize);
+
+    bufferFile = createBufferFile("stream-buffer");
+    try {
+      fileStore = new RandomAccessFile(bufferFile, "rw");
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException(format("Buffer file %s was just created but now it doesn't exist",
```
marianogonzalez: yes

src/main/java/org/mule/runtime/internal/stream/bytes/FileStoreTraversableBuffer.java
```diff
@@ -0,0 +1,266 @@
...
+        ByteBuffer src = buffer.duplicate();
+        final int newPosition = toIntExact(requiredRange.start - bufferRange.start);
+        src.position(newPosition);
+        src.limit(newPosition + min(length, bufferSize - newPosition));
+        dest.put(src);
+        return length;
+      }
+
+      if (bufferRange.isAhead(requiredRange) || streamFullyConsumed) {
+        bufferLock.unlock(); // we don't need to hold this lock anymore
```
marianogonzalez: no. In this case I know I adquired it

src/main/java/org/mule/runtime/internal/stream/bytes/TraversableBuffer.java
```diff
@@ -0,0 +1,34 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.internal.stream.bytes;
+
+import java.nio.ByteBuffer;
+
+/**
+ * A buffer which provides concurrent random access to the entirety
```
marianogonzalez: nop.

src/main/java/org/mule/runtime/internal/stream/bytes/TempFileHelper.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.concurrent.ExecutorService;
+
+final class TempFileHelper {
+
+  private static final String TEMP_DIR_SYSTEM_PROPERTY = "java.io.tmpdir";
+  private static final File TEMP_DIR = new File(System.getProperty(TEMP_DIR_SYSTEM_PROPERTY));
+  private static Random RANDOM = new SecureRandom();
+
+  //TODO: this cannot be static
+  private static final ExecutorService asyncService = newFixedThreadPool(1);
```
marianogonzalez: as the jdoc says, this is temporal and needs to be removed

PR #146, opened 2017-01-24
MULE-11529: Define interception API

elrodro83:


src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+   * @param event an object that contains the state of the event to be sent to the processor. It may be modified by calling its
+   *        mutator methods.
+   * @param action when something other than continuing the interception is desired, the corresponding method on this object must
+   *        be called.
+   */
+  void before(String flowPath, Map<String, Object> parameters, InterceptionEvent event, InterceptionAction action);
+
+  /**
+   * Used only for notification, this method will be called after the intercepted processor has run, or after it was skipped in
+   * the {@link #before(String, Map, InterceptionEvent, InterceptionAction) before} method.
```
gsfernandes: Perhaps it should also say if the intercepted processor throws an error, this method still will be called.

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
```diff
@@ -0,0 +1,73 @@
...
+import org.mule.runtime.api.metadata.DataType;
+
+import java.util.Map;
+
+/**
+ * Extends {@link InterceptionEventResult} adding mutator methods.
+ *
+ * @since 1.0
+ */
+public interface InterceptionEvent extends InterceptionEventResult {
```
dfeist: Can any of this be made more generic so it can be used elsewhere such as in scripting components? Even if this is not done now, then is it worth adding a TODO?

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.api.interception;
+
+import java.util.Map;
+
+/**
+ * Provides a way to hook behavior around a Processor. Implementations must implement the
+ * {@link #before(String, Map, InterceptionEvent, InterceptionAction) before} method and may or may not implement
```
elrodro83: Will change before to be a default empty method.

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ * A processor may have more than one interceptor applied. In that case, all will be applied in a predetermined order, calling the
+ * {@link #before(String, Map, InterceptionEvent, InterceptionAction) before} methods of each, then the processor itself and
+ * finally the {@link #after(InterceptionEventResult) after} methods in inverse order.
+ * <p>
+ * When initialized, all fields in the implementation with @{@link javax.inject.Inject} will be set with an object obtained from
+ * the registry by type.
+ * 
+ * @since 1.0
+ */
+public interface InterceptionHandler {
```
elrodro83: We reviewed the use cases with the users of this API, and found no need to provide 2 different kind of interceptors. Also, keep in mind that the users will also have to register these handlers, and having different typed will make more complex he usage from their side.

The `around` as you describe it would cause the effect of non-skippable processors being skipped, which is a case we want to avoid. This came as a concern when we reviewed this API with the potential users.

What performance impact do you mean? there is no flatMap here, this is only an interface, and on the implementation done on top of this api there is no `flatMap`.

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * This method is called before the intercepted processor has run. It may call a method on {@code action} to do anything other
+   * than continue with the interception chain, modify the event to be used down in the chain and the processor, or both.
+   * 
+   * @param flowPath the path of the intercepted processor in the mule app configuration.
```
elrodro83: Makes sense to separate the `before`and a predicate. Will analyze this further with @gsfernandes

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+   */
+  void before(String flowPath, Map<String, Object> parameters, InterceptionEvent event, InterceptionAction action);
+
+  /**
+   * Used only for notification, this method will be called after the intercepted processor has run, or after it was skipped in
+   * the {@link #before(String, Map, InterceptionEvent, InterceptionAction) before} method.
+   * 
+   * @param event the result of the processor.
+   */
+  default void after(InterceptionEventResult event) {}
```
elrodro83: the flowPath and parameters are used to determine the action to take. Those may be moved to a predicate.

src/main/java/org/mule/runtime/api/interception/InterceptionAction.java
```diff
@@ -0,0 +1,36 @@
...
+ * @since 1.0
+ */
+public interface InterceptionAction {
+
+  /**
+   * Only operation processors or similar may be skipped. Routing or intercepting processors may not be intercepted.
+   * 
+   * @return {@code true} if the intercepted processor may be {@link #skip() skipped}.
+   */
+  boolean isSkippable();
```
elrodro83: some processors may be intercepted but not skipped.

src/main/java/org/mule/runtime/api/interception/InterceptionAction.java
```diff
@@ -0,0 +1,29 @@
...
+  /**
+   * Interrupts the current interception chain, effectively skipping the remaining {@link InterceptionHandler handlers} and the
+   * intercepted processor.
+   * <p>
+   * In the case an {@link InterceptionHandler} calls this method, the following handlers in the chain nor the processor will be
+   * run. Only the {@link InterceptionHandler#after(InterceptionEventResult)}s for the calling {@link InterceptionHandler handler}
+   * and those for which {@link InterceptionHandler#before(String, java.util.Map, InterceptionEvent, InterceptionAction)} was
+   * called will be invoked.
+   */
+  void skip();
```
elrodro83: can you clarify this issue?

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
elrodro83: making this more generic seems like a way to get the event in the API again. I think it would be better for each api thar requires it to have its taylored view of the event.

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,86 @@
...
+   * continue and eventually call the intercepted component.
+   * 
+   * @param parameters the parameters of the component as defined in the configuration. Parameters that contain expressions will
+   *        be resolved when passed to this method.
+   * @param event an object that contains the state of the event to be sent to the component. It may be modified by calling its
+   *        mutator methods.
+   * @param action when something other than continuing the interception is desired, the corresponding method on this object must
+   *        be called.
+   */
+  default void around(Map<String, Object> parameters, InterceptionEvent event, InterceptionAction action) {
```
dfeist: Doesn't this need access to the `after` event?  You could argue that this can be used along with the implementation of an `after` method but that's not the same as you can only achieve that use cases that need both via state sharing.

As it is it's more `next` than `around`, with the ability to do things after next and access to event before/after it's an `around`

src/main/java/org/mule/runtime/api/interception/InterceptionAction.java
```diff
@@ -0,0 +1,41 @@
...
+ * 
+ * @since 1.0
+ */
+public interface InterceptionAction {
+
+  /**
+   * Indicates that the current interception chain must continue, proceeding with the next {@link InterceptionHandler handlers} in
+   * the chain (if any) and the intercepted component.
+   * 
+   * @return a {@link CompletableFuture} with the {@link InterceptionEvent event} that results of the intercepted processor.
```
dfeist: It is the result of executing the rest of the chains (not just the processor)

src/main/java/org/mule/runtime/api/interception/InterceptionAction.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Interrupts the current interception chain, skipping the next {@link InterceptionHandler interceptors} in the chain and the
+   * intercepted component.
+   * <p>
+   * If called for a <b>router</b> component, the whole route (and all the components contained in it) will be skipped.
+   * <p>
+   * If called for an intercepting component (i.e. a {@code splitter}), the listener components (i.e. what follows the
+   * {@code splitter} up to an {@code aggregator}) are skipped as well.
+   * 
+   * @return a {@link CompletableFuture} with the same {@link InterceptionEvent event} as it was at the time this method was
```
dfeist: Worth explicitly saying this is equivalent to `java.util.concurrent.CompletableFuture#completedFuture()` and here for usability? Make slightly clearer -> "using the mutable event passed to the around method"?

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,102 @@
...
+   * If implemented, only by calling {@code action} {@link InterceptionAction#proceed() proceed()} will the interception chain
+   * continue and eventually call the intercepted component. Otherwise, by calling {@link InterceptionAction#skip() skip()} the
+   * interception chain execution will be interrupted and {@link #after(InterceptionEvent) after} method called immediately.
+   * ({@link InterceptionAction#skip() skip()} may not be called at all, but it is convenient that it already returns a
+   * {@link CompletableFuture} to return in this method.)
+   * <p>
+   * Calling an implementation with this method will be less efficient than calling just {@link #before(Map, InterceptionEvent)
+   * before} and {@link #after(InterceptionEvent) after}. So, {@link #around(Map, InterceptionEvent, InterceptionAction) around}
+   * should only be implemented for cases that cannot be done just with {@link #before(Map, InterceptionEvent) before} and/or
+   * {@link #after(InterceptionEvent) after}.
```
dfeist: For example if you need to i) dynamically determine if the next of the interceptor chains should be processed or not ii) need to perform non-blocking operations before/after ...

src/main/java/org/mule/runtime/api/interception/InterceptionHandlerFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Factory for concrete {@link InterceptionHandler} objects.
+ * <p>
+ * When initialized, all fields in the implementation with @{@link javax.inject.Inject} will be set with an object obtained from
+ * the registry by type.
+ * 
+ * @since 1.0
+ */
+public interface InterceptionHandlerFactory {
```
dfeist: Not sure why this is needed, but i'll look at implementation first to understand...

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
```diff
@@ -0,0 +1,102 @@
...
+ * Interceptable components are those that are defined by a configuration element and have a {@link ComponentLocation}.
+ * <p>
+ * A component may have more than one interceptor applied. In that case, all will be applied in a predetermined order, calling the
+ * {@link #before(Map, InterceptionEvent) before} methods of each, {@link #around(Map, InterceptionEvent, InterceptionAction)
+ * around} methods which may proceed to the next interceptor or the component itself and finally the
+ * {@link #after(InterceptionEvent) after} methods.
+ *
+ * @since 1.0
+ */
+public interface InterceptionHandler {
```
dfeist: This name is fine, just wondering why not `Interceptor` or `ProcessorInterceptor` directly wasn't used.

src/main/java/org/mule/runtime/api/interception/InterceptionAction.java
elrodro83: saying it calls `completedFutire` is exposing an implementation detail. I don think that needs to be metioned.

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
elrodro83: good question :P. we just thought of it this way at the beginning.,

elrodro83:
@dfeist what do you think of changing `after(InterceptionEvent event)` to `after(Optional<InterceptionEvent> event, Optional<Throwable> thrown)`?

src/main/java/org/mule/runtime/api/interception/InterceptionHandlerFactory.java
elrodro83: this may just be needed when the registration part is done, not yet.

src/main/java/org/mule/runtime/api/interception/InterceptionEventResult.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Provides access to the attributes of the input event of a component.
+ *
+ * @since 1.0
+ */
+public interface InterceptionEventResult {
```
dfeist: Why is this still here, I don't see it being used.

src/main/java/org/mule/runtime/api/interception/InterceptionHandler.java
dfeist: Worth changing to `ProcessorInterceptor`, it would be clear what's being intercepted that way..

src/main/java/org/mule/runtime/api/interception/InterceptionHandlerFactory.java
dfeist: So can it be removed from this PR and only added if/when needed? I personally don't see why a factory should be needed.

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
dfeist: Thinking about it, we probably don't want to give components access to Event anyway, rather a mutable message, so there is no other potential use for this currently.

src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Unique identifier for a configuration option. Every configuration option has a namespace and an identifier.
+ * <p>
+ * The namespace is a short name of the extension that defines the component. Even core configuration have a namespace even though
+ * they have namespace in the declaration in the configuration files.
+ *
+ * @since 1.0
+ */
+public interface ComponentIdentifier {
```
dfeist: `component` is a very generic term, wondering if there is anything more specific. I guess the issue with `element` is that it's somewhat XML specific..

dfeist: `component` is a very generic term, wondering if there is anything more specific. I guess the issue with `element` is that it's somewhat XML specific..

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  InterceptionEvent removeVariable(String key);
+
+  /**
+   * Sets an error related to the produced event.
+   *
+   * @param error the error associated with the produced event
+   * @return the builder instance
+   */
+  InterceptionEvent error(Error error);
```
dfeist: Does this support null?

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
```diff
@@ -0,0 +1,72 @@
...
+   * @param variables variables to be set.
+   * @return the builder instance
+   */
+  InterceptionEvent variables(Map<String, Object> variables);
+
+  /**
+   * Add a variable.
+   *
+   * @param key the key of the variable to add.
+   * @param value the value of the variable to add.
```
dfeist: Are null values supported?

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
```diff
@@ -0,0 +1,72 @@
...
+   * @return the builder instance.
+   */
+  InterceptionEvent addVariable(String key, Object value, DataType mediaType);
```
dfeist: Why `DataType` here, and not `mediaType` directly?  Also is a method that takes `TypedValue` preferred here to match getter?

src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
elrodro83: we came up with the term *component* as something that can be a Processor or an operation from an API point of view.
What this comment does is expose the need we have for a glossary though.

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
elrodro83: As it turns out, this isn't needed. The proper way of getting an error to be set is by throwing an exception.

src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Regular component that doesn't match any of the other criterias.
+     */
+    PROCESSOR,
+
+    /**
+     * Identifies an {@link Error} thrown by another component.
+     */
+    ERROR;
```
dfeist: Shouldn't this be ERROR_HANLER? How can ERROR be a component type. @pablolagreca

src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Executes an operation defined in an extension.
+     */
+    OPERATION,
+
+    /**
+     * Executes one or many nested component chains.
+     */
+    ROUTER,
```
dfeist: Scope?  Or are all scopes routers? (they aren't in studio)

elrodro83: that's a discussion for another moment i think

PR #150, opened 2017-01-26
MULE-11540: Have an ExpressionManager interface in API

elrodro83:


src/main/java/org/mule/runtime/api/el/ExpressionEvaluator.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Evaluates an expression according to a given {@link BindingContext} and the global one.
+   *
+   * @param expression the EL expression
+   * @param context the current dynamic binding context to consider
+   * @return the result of the expression plus its type
+   * @throws ExpressionExecutionException when an error occurs during evaluation
+   */
+  TypedValue evaluate(String expression, BindingContext context) throws ExpressionExecutionException;
```
afelisatti: We could also add the method that uses the desired output DataType.

PR #155, opened 2017-02-03
MULE-11482 - Reimplement flow path mechanism so it does not depend on…

pablolagreca:
… each processor

- Add new location mechanism
- Pending remove old code for setting the processorPath in all the components
- More commits will be done under the same issue to complete the task

### DIFF
--- a/src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
+++ b/src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
@@ -6,7 +6,7 @@
  */
 package org.mule.runtime.api.component;
 
-import org.mule.runtime.api.message.Message;
+import static org.mule.runtime.api.component.DefaultComponentIdentifier.parseComponentIdentifier;
 
 /**
  * Unique identifier for a configuration option. Every configuration option has a namespace and an identifier.
@@ -17,36 +17,6 @@ import org.mule.runtime.api.message.Message;
  * @since 1.0
  */
 public interface ComponentIdentifier {
-
-  /**
-   * Declares different types of component based on its common characteristics.
-   */
-  public enum ComponentType {
-    /**
-     * Receives something from an external system, transforms it into a {@link Message} and vice-versa.
-     */
-    SOURCE,
-
-    /**
-     * Executes an operation defined in an extension.
-     */
-    OPERATION,
-
-    /**
-     * Executes one or many nested component chains.
-     */
-    ROUTER,
-
-    /**
-     * Wraps the next defined component, controlling its invocation.
-     */
-    INTERCEPTING,
-
-    /**
-     * Regular component that doesn't match any of the other criterias.
-     */
-    PROCESSOR;
-  }
 
   /**
    * The namespace is a short name of the extension that defines the component.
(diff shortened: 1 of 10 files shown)


src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
```diff
@@ -61,7 +31,48 @@
...
+   * @since 1.0
    */
-  ComponentType getComponentType();
+  interface Builder {
+
+    /**
+     * @param name name of the component
+     * @return {@code this} builder
+     */
+    Builder withName(String name);
```
dfeist: The use of "with" doesn't match other builders that don't use it. Ideally we should be consistent with builders, at least those in same module.

src/main/java/org/mule/runtime/api/component/DefaultComponentIdentifier.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Unique identifier for a configuration option. Every configuration option has a namespace and an identifier.
+ *
+ * The namespace define the extension that defines the component. Even core configuration has a namespace even though there's no
+ * namespace used in the configuration files.
+ *
+ * @since 4.0
+ */
+class DefaultComponentIdentifier implements ComponentIdentifier {
```
dfeist: Elsewhere implementation is private/hidden inside builder, rather than this way around. Any just consistency..

src/main/java/org/mule/runtime/api/component/DefaultTypedComponentIdentifier.java
```diff
@@ -0,0 +1,72 @@
...
+  public ComponentIdentifier getIdentifier() {
+    return identifier;
+  }
+
+  public ComponentType getType() {
+    return type;
+  }
+
+  public static class Builder implements TypedComponentIdentifier.Builder {
+
```
dfeist: Same comments apply to this builder.

src/main/java/org/mule/runtime/api/component/TypedComponentIdentifier.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Wraps the next defined component, controlling its invocation.
+     */
+    INTERCEPTING,
+
+    /**
+     * Regular component that doesn't match any of the other criterias.
+     */
+    PROCESSOR,
```
dfeist: What category do scopes use? Router or processor or unknown?

src/main/java/org/mule/runtime/api/component/TypedComponentIdentifier.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Error handler on error child component type
+     */
+    ON_ERROR,
+
+    /**
+     * Component type for errors identifiers.
+     */
+    ERROR,
```
dfeist: How is this a component type?

src/main/java/org/mule/runtime/api/meta/AbstractAnnotatedObject.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.xml.namespace.QName;
+
+/**
+ * Base implementation for {@link AnnotatedObject}
+ */
+public abstract class AbstractAnnotatedObject implements AnnotatedObject {
```
dfeist: Does abstract implementation really need to be in API when interface is in API?

src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
pablolagreca: I was following other builder. I have another commit over the same issue. Will remove the "with" in those commits.

src/main/java/org/mule/runtime/api/component/TypedComponentIdentifier.java
pablolagreca: I just keeping what was introduced for Interception API here. I Just move this.

src/main/java/org/mule/runtime/api/component/TypedComponentIdentifier.java
pablolagreca: It's not. I will remove it.

src/main/java/org/mule/runtime/api/meta/AbstractAnnotatedObject.java
pablolagreca: It's a helper class that can enforce correct behaviour for all those client that can inherit from it.

src/main/java/org/mule/runtime/api/component/ComponentIdentifier.java
dfeist: Take a look at Message/DataType and those other interfaces.  i) no `with`, ii) builder exposed in interface via builder() iii) impl hidden in builder rather than it's own visible class with private constructor.

src/main/java/org/mule/runtime/api/component/TypedComponentIdentifier.java
dfeist: I know, i asked Rodo this but he said to ask you in your PR that was coming next, because you came up with the list :-D

PR #159, opened 2017-02-08
MULE-11284: Improve registration of global contexts

afelisatti:


src/main/java/org/mule/runtime/api/el/ExpressionEvaluator.java
```diff
@@ -18,6 +18,14 @@
 public interface ExpressionEvaluator {
 
   /**
+   * Registers the given {@link BindingContext} entries as globals. Notice globals cannot be removed once registered, only
```
elrodro83: entry, in singular

afelisatti: I'm referring to the entries within the context.

PR #167, opened 2017-03-03
DO NOT MERGE: REVIEW ONLY - MULE-11839: Support core components in the Extension Model

marianogonzalez:


src/main/java/org/mule/runtime/api/meta/model/Stereotype.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.mule.runtime.api.meta.model;
+
+import static org.mule.runtime.api.util.Preconditions.checkArgument;
+import org.mule.runtime.api.meta.NamedObject;
+
+import java.util.Objects;
+
+/**
+ * A widely held but fixed and oversimplified image or idea of the owning model
```
pablolagreca: Can you provide an example as part of the javadoc?.

src/main/java/org/mule/runtime/api/meta/model/operation/RouteModel.java
```diff
@@ -0,0 +1,65 @@
...
+ * </ul>
+ * <p>
+ * Each of the above constitute a {@code Route}. Several instances of those routes can exists
+ * inside the same component instance. For example, the {@code when} route can be used many times
+ * inside a choice router. On the other hand, the {@code otherwise} route inside the same router can
+ * only exist up to one time, while it is also allowed to not exist at all. This {@link #getMinOccurs()}
+ * and {@link #getMaxOccurs()} methods model that.
+ * <p>
+ * Finally, some scope have restrictions regarding which components can be used with them. For example,
+ * {@code validation:all} only allows other validators to be placed inside. That's what the {@link #getAllowedStereotypes()}
```
pablolagreca: If we use the concept "Processor" as an abstraction of Router, Scope, Operation, then we should call this something link getAllowedProcessorsType(). Then a "Processor" will have a particular type and this method will return the "types" the "Processors" must comply to.

src/main/java/org/mule/runtime/api/meta/model/operation/RouterModel.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * An {@link OperationModel} specification which represents the particular case of a
+ * router, such as choice, scatter-gather, first-successful, etc.
+ * <p>
+ * It adds a {@link #getRouteModels() list of routes} which describe the contained routes.
+ *
+ * @since 1.0
+ */
+public interface RouterModel extends OperationModel {
```
pablolagreca: Same as with the *Declarer classes. We need to create an abstraction like "Processor". See extensions-api PR comment.

src/main/java/org/mule/runtime/api/meta/model/operation/RouteModel.java
marianogonzalez: There're cons to doing what you suggest. We should discuss it off line and if we decide to take this route (pun intended) we can do it as part of MULE-11912

PR #170, opened 2017-03-07
MULE-11936: IllegalArgumentException: Can not set final java.lang.Str…

gsfernandes:
…ing field org.mule.runtime.api.metadata.DefaultMetadataKey.id to org.mule.runtime.extension.api.metadata.NullMetadataKey

### DIFF
--- a/src/main/java/org/mule/runtime/api/metadata/DefaultMetadataKey.java
+++ b/src/main/java/org/mule/runtime/api/metadata/DefaultMetadataKey.java
@@ -24,7 +24,7 @@ import java.util.Set;
  *
  * @since 1.0
  */
-public final class DefaultMetadataKey implements MetadataKey {
+public class DefaultMetadataKey implements MetadataKey {
 
   private final String id;
   private final String displayName;


src/main/java/org/mule/runtime/api/metadata/DefaultMetadataKey.java
```diff
@@ -24,7 +24,7 @@
  *
  * @since 1.0
  */
-public final class DefaultMetadataKey implements MetadataKey {
+public class DefaultMetadataKey implements MetadataKey {
```
estebanwasinger: I don't think that this is the correct solution, `NullMetadataKey` problems should not impact `DefaultMetadataKey`

gsfernandes: I know, I didn't wanted to make this change neither. The other approach would be to have a "hierarchy" adapter that would need to know when to create a NullMetadataKey or DefaultMetadataKey depending on a "field" type for instance or we could just check for id == "". @juandesi, would you agree on this?

juandesi: I don't think this is a big deal, I agree with esteban we could find a better solution, If is a quick fix on the adapter side I would go with that, but still don't think this is a big problem.

PR #179, opened 2017-03-27
MULE-12010: Bundle import/export of declaration in a single Artifact …

alegmarra:
…serializer


src/main/java/org/mule/runtime/api/app/declaration/fluent/ParameterListValue.java
```diff
@@ -85,4 +86,8 @@ public int hashCode() {
     return values.hashCode();
   }
 
+  @Override
+  public String toString() {
+    return "[" + values.stream().map(Object::toString).collect(joining(", ")) + "]";
```
marianogonzalez: doesn't the toString of a list already does this?

PR #180, opened 2017-03-28
MULE-11221: Support Transactions in Sources

estebanwasinger:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ComponentDeclarer.java
```diff
@@ -14,7 +14,7 @@
  *
  * @since 1.0
  */
-abstract class ComponentDeclarer<T extends ComponentDeclarer, D extends ComponentDeclaration>
+public abstract class ComponentDeclarer<T extends ComponentDeclarer, D extends ComponentDeclaration>
```
estebanwasinger: required for --> https://github.com/mulesoft/mule/pull/4370/files#diff-2fd54bb05b1bae4719fcfd253350afe6R64

PR #183, opened 2017-03-29
MULE-11837: Simplify declaration to make everything a parameter

alegmarra:
@marianogonzalez  @pablolagreca, I've created this early PR so you can check the new ArtifactDeclaration API. 

There's still work to do in mule core and the tooling before merging, but I don't think that will affect this API, and I'd appreciate early feedback since the schedule is tight.


marianogonzalez:
LGTM

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Adds a {@link ParameterElementDeclaration custom parameter} to {@code this} enrichable element declaration.
+   * This {@code customParameter} represents an additional parameter to the ones exposed by the actual model
+   * associated to this {@link ElementDeclaration element}.
+   * No validation of any kind will be performed over this {@code customParameter} and its value.
+   *
+   * @param customParameter the {@link ParameterElementDeclaration} to associate to {@code this} element declaration
+   */
+  void addCustomParameter(ParameterElementDeclaration customParameter);
```
pablolagreca: Can you put an example of the purpose of custom parameters?

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
```diff
@@ -0,0 +1,55 @@
...
+
+
+  /**
+   * Adds a property to the {@link ElementDeclaration}.
+   * This property is meant to hold only metadata of the declaration.
+   *
+   * @param name custom attribute name.
+   * @param value custom attribute value.
+   */
+  void addProperty(String name, Object value);
```
pablolagreca: What is the purpose of properties? Can you put an example?

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
```diff
@@ -0,0 +1,55 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.app.declaration;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * //TODO
```
pablolagreca: Important for this review

src/main/java/org/mule/runtime/api/app/declaration/ArtifactDeclaration.java
```diff
@@ -18,62 +18,26 @@
...
-  private List<TopLevelParameterDeclaration> globalParameters = new LinkedList<>();
-  private List<ConfigurationElementDeclaration> configs = new LinkedList<>();
-  private List<FlowElementDeclaration> flows = new LinkedList<>();
+  private List<GlobalElementDeclaration> globalElements = new LinkedList<>();
 
   public ArtifactDeclaration() {}
 
   /**
-   * @return the {@link List} of {@link FlowElementDeclaration flows} associated with
+   * @return the {@link List} of {@link GlobalElementDeclaration global elements} associated with
```
pablolagreca: 👍

src/main/java/org/mule/runtime/api/app/declaration/FlowElementDeclaration.java
```diff
@@ -6,57 +6,24 @@
...
 /**
  * A programmatic descriptor of an application Flow configuration.
  *
  * @since 1.0
  */
-public final class FlowElementDeclaration extends ParameterizedElementDeclaration {
+public final class FlowElementDeclaration extends ScopeElementDeclaration implements GlobalElementDeclaration {
 
-  private List<ComponentElementDeclaration> components = new LinkedList<>();
+  private static final String CE_EXTENSION_NAME = "Mule Core";
```
pablolagreca: I think this should be the namespace we use for core components which is "mule"

src/test/java/org/mule/runtime/api/meta/model/DeclarationSerializationTestCase.java
```diff
@@ -59,15 +59,15 @@ private ArtifactDeclaration createArtifact() {
     ElementDeclarer core = ElementDeclarer.forExtension("Mule Core");
 
     return newArtifact()
-        .withGlobalParameter(db.newGlobalParameter("query")
+        .withGlobalElement(db.newGlobalParameter("query")
             .withRefName("selectQuery")
```
pablolagreca: Not part of this PR but I don't think withRefName is the best name for the method. Isn't this the name parameters?

src/test/resources/declaration/artifact-declaration.json
```diff
@@ -1,245 +1,270 @@
...
-      "elementName": "selectQuery",
-      "value": {
-        "parameters": {
-          "sql": "select * from PLANET where name \u003d :name",
-          "inputParameters": "#[mel:[\u0027name\u0027 : payload]]"
-        },
-        "typeId": "org.mule.extension.db.api.param.QueryDefinition"
-      },
+      "name": "query",
+      "declaringExtension": "Database",
```
pablolagreca: Can we call this "namespace" ?

src/test/resources/declaration/artifact-declaration.json
```diff
@@ -1,245 +1,270 @@
...
+      "kind": "GLOBAL_PARAMETER",
       "customParameters": [],
       "properties": {},
-      "name": "query",
-      "declaringExtension": "Database"
-    }
-  ],
-  "configs": [
+      "refName": "selectQuery",
```
pablolagreca: why "refName"?

alegmarra: Because is the name we'd use in the `config-ref` parameter of an operation. 
If we have `<http:request-config name="reqConfig">`
So `name` is reserved for the model-name of the element `request-config` and we need some way to declare the configured "referable" name `reqConfig`

src/test/java/org/mule/runtime/api/meta/model/DeclarationSerializationTestCase.java
alegmarra: yes, we can use `name` here, since the "model name" is passed when creating the element. See my other comment regarding `ref-name` vs `model-name`

src/main/java/org/mule/runtime/api/app/declaration/FlowElementDeclaration.java
alegmarra: right now, every time we need to find an extension in a given context internally we are using the extension name as an id, and the Core extension model has namespace `mule` but name `Mule Core`. 
I don't think is ideal replacing the name of the core extension with `mule`. In the other hand, using the namespace as identifier is something yet to be defined in the bigger picture, and if we want to be consistent in the meantime, we have to use the extension name here.

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
alegmarra: Properties are more of an internal element, but necessary for handling metadata over a given element that was declared. 
For example, if we load an artifact that had declared custom prefixes and they use `my-jms:consume` insead of `jms:consume`, we need to have `jms` as the namespace of the declaration, but we are also required to honour the `my-jms` custom prefix upon serialization.
In that case, we have a `CUSTOM_PREFIX` property, that can be propagated and used when we need it.

alegmarra: glad to hear better ideas on how to solve this, adding "properties" metadata is far from ideal.

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
alegmarra: There are some cases where we need to keep parameters that are not present in the model of a given element.
Since we need to parse full artifacts now, and it is expected that for a given xml that is parsed we generate an identical app, we need to handle special attributes that are not part of the model. 
For example, `mule` ArtifactDeclaration has xml attributes that are important to keep in load/serialize scenarios, like xmlns alias definitions.
Other case are the `doc:name` attributes that are added to an operation. There is nothing in the model that relates to that, but we need to handle them anyway.

marianogonzalez: I'm not against the concept of custom parameters, but it calls my attention that the examples you mention are pretty core things. Be careful not to repeat the mistakes made with things like ModelProperties or TypeAnnotations. These things sound core enough to be explicit in the model. That being said, I'm sure there're other examples of things we do probably want to treat as custom parameters

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
marianogonzalez: same comment as before. A CUSTOM_PREFIX property is useless unless there's somebody on the "other side of the house" who is expecting that property. If the property is known to the model clients, then why not just putting it in the model?

Maybe you're choosing your examples wrong, but it does sound like a bad case of `cat bag` pattern. Also, it seems to be a big overlap between properties and custom parameters

src/test/resources/declaration/artifact-declaration.json
alegmarra: this is really the extension Name (yeah, should have been `declaringExtensionName`), calling it `namespace` might be confusing. When we have an extension ID that is common from tooling to runtime then this field would be that id.

src/main/java/org/mule/runtime/api/app/declaration/FlowElementDeclaration.java
pablolagreca: We need to reconcile what the extension name and the namespace name is. I think that the concept of namespace represents much better the purpose we are using the extension name for. Let's discuss it offline

src/test/resources/declaration/artifact-declaration.json
pablolagreca: Extension ID == namespace I think.

src/main/java/org/mule/runtime/api/app/declaration/EnrichableDeclaration.java
```diff
@@ -14,50 +15,78 @@
...
 
   /**
-   * Adds a property to the {@link ElementDeclaration}.
-   * This property is meant to hold only metadata of the declaration.
+   * Adds a {@code metadataProperty} to the {@link ElementDeclaration}.
+   * This property is meant to hold only metadata of the declaration,
+   * related to how the declaration has to be represented but not affecting nor containing
+   * information related to the model configured with this declaration.
+   *
+   * This property may contain information regarding things like transformations required for
```
pablolagreca: Can you add an example? Sorry for being a PITA about this but this API is going to be critical for all tooling clients.

PR #190, opened 2017-04-03
MULE-12001: ArtifactDeclarationLoader support for EE namespace

pablomusumeci:


src/main/java/org/mule/runtime/internal/dsl/DslConstants.java
```diff
@@ -33,6 +33,16 @@
   String CORE_NAMESPACE = format(DEFAULT_NAMESPACE_URI_MASK, "core");
 
   /**
+   * This is the namespace prefix for EE elements in the configuration.
+   */
+  String EE_PREFIX = "ee";
+
+  /**
+   * Location of Mule's EE schema
```
alegmarra: Is not the location of the schema, it's the namespace for EE elements. 
Fix this doc in the "core prefix" too, please

PR #192, opened 2017-04-04
MULE-12140 Modifing api to support expected output and split

machaval:


src/main/java/org/mule/runtime/api/el/MuleExpressionLanguage.java
```diff
@@ -0,0 +1,45 @@
...
+    return validate(expression).isSuccess();
+  }
+
+  /**
+   * Verifies whether an expression is valid or not syntactically.
+   *
+   * @param expression to be validated
+   * @return a {@link ValidationResult} indicating whether the validation was successful or not
+   */
+  ValidationResult validate(String expression);
```
afelisatti: Why is validate repeated here and in the language?

src/main/java/org/mule/runtime/api/el/DefaultExpressionLanguageFactoryService.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.mule.runtime.api.el;
+
+
+import org.mule.runtime.api.service.Service;
+
+/**
+ * Provides a factory for the default Expression language that is going to be used by the {@link MuleExpressionLanguage}
+ */
+public interface DefaultExpressionLanguageFactoryService extends Service {
```
afelisatti: Why is this the "default"? Isn't just to create DW?

machaval: It is not harcoded to DW. We can replace the service from DW to groovy and it will be the same

PR #193, opened 2017-04-05
MULE-11767: Improve TLS API

afelisatti:


src/main/java/org/mule/runtime/api/tls/AbstractTlsContextFactoryBuilderFactory.java
```diff
@@ -0,0 +1,59 @@
...
+  static {
+    try {
+      final AbstractTlsContextFactoryBuilderFactory factory =
+          load(AbstractTlsContextFactoryBuilderFactory.class).iterator().next();
+      LOGGER.info(format("Loaded TlsContextFactoryBuilderFactory implementation '%s' form classloader '%s'",
+                         factory.getClass().getName(), factory.getClass().getClassLoader().toString()));
+
+      DEFAULT_FACTORY = factory;
+    } catch (Throwable t) {
+      LOGGER.error("Error loading TlsContextFactoryBuilderFactory implementation.", t);
```
pablolagreca: I would add something like "seems the api implementation is not present in the classpath"

src/main/java/org/mule/runtime/api/tls/TlsContextFactory.java
```diff
@@ -24,13 +26,47 @@
...
+   * Provides a builder to create custom {@link TlsContextFactory} objects or obtain the default one.
+   *
+   * @return a new {@link TlsContextFactoryBuilder}
+   */
+  static TlsContextFactoryBuilder builder() {
+    return AbstractTlsContextFactoryBuilderFactory.getDefaultFactory().create();
+  }
+
+  /**
+   * Allows the creation of an {@link SSLContext} with the configured keystore and trust store. Use {@link #getEnabledProtocols()}
```
pablolagreca: "Use" -> "You must use" ?

PR #197, opened 2017-04-07
MULE-12061 Remove the RouteModel from ScopeModel

alegmarra:


src/main/java/org/mule/runtime/api/meta/model/operation/ScopeModel.java
```diff
@@ -6,22 +6,45 @@
...
  */
 public interface ScopeModel extends OperationModel {
 
   /**
-   * @return The model of the owned route
+   * Represents the minimum amount of times that this route can be used inside the owning component.
+   *
+   * @return An int greater or equal to zero
+   */
+  int getMinOccurs();
```
marianogonzalez: this property only makes sense when the scope is being used as a route. It doesn't have any sense on a stand alone scope

src/main/java/org/mule/runtime/api/meta/model/operation/ScopeModel.java
```diff
@@ -6,22 +6,45 @@
...
+
+  /**
+   * {@link Optional} value which represents the maximum amount of times that this route can be used inside the owning
+   * component.
+   *
+   * @return If present, a number greater or equal to zero.
+   */
+  Optional<Integer> getMaxOccurs();
```
marianogonzalez: idem

src/main/java/org/mule/runtime/api/meta/model/operation/ScopeModel.java
```diff
@@ -6,22 +6,45 @@
...
+  /**
+   * {@link Optional} {@link Set} of {@link Stereotype stereotypes} to which the components placed inside the route
+   * <b>MUST</b> conform to.
+   * <p>
+   * If the value is empty, it means that no restriction applies.
+   *
+   * @return If present, a non empty set
    */
-  RouteModel getRouteModel();
+  Optional<Set<Stereotype>> getAllowedStereotypes();
```
marianogonzalez: idem

PR #210, opened 2017-04-25
MULE-10517: Add information on the EventContext about the origin of its events

afelisatti:


src/main/java/org/mule/runtime/api/component/DefaultTypedComponentIdentifier.java
```diff
@@ -8,7 +8,9 @@
 
 import static org.mule.runtime.api.util.Preconditions.checkState;
 
-class DefaultTypedComponentIdentifier implements TypedComponentIdentifier {
+import java.io.Serializable;
+
+class DefaultTypedComponentIdentifier implements TypedComponentIdentifier, Serializable {
```
elrodro83: add serial version id

src/main/java/org/mule/runtime/api/component/DefaultComponentIdentifier.java
```diff
@@ -19,7 +21,7 @@
  *
  * @since 4.0
  */
-class DefaultComponentIdentifier implements ComponentIdentifier {
+class DefaultComponentIdentifier implements ComponentIdentifier, Serializable {
```
elrodro83: add serial version id

PR #214, opened 2017-05-03
MULE-11749: handle fatal JVM errors so stream processing continues

alepulver:
DO NOT MERGE YET

### DIFF
--- a/src/main/java/org/mule/runtime/api/exception/MuleFatalJvmException.java
+++ b/src/main/java/org/mule/runtime/api/exception/MuleFatalJvmException.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/exception/MuleFatalJvmException.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.mule.runtime.api.i18n.I18nMessage;
+
+/**
+ * {@code MuleFatalJvmException} Is the runtime exception used to wrap fatal errors such as {@code VirtualMachineError}. This
+ * is to avoid Reactor intercepting them in a stream and propagating outside to arbitrary threads.
+ *
+ * @since 4.0
+ */
+public class MuleFatalJvmException extends RuntimeException {
```
dfeist: -Is this really needed in API?
- Why doesn't this extend MuleRuntimeException?
- Do we need to mention JVM, can't we use `MuleFatalException`

alepulver: Applied suggestions, and moved to core.

dfeist: So this PR should be closed now?

PR #224, opened 2017-05-17
MULE-12434: Poll extension model is invalid

pablomusumeci:


src/main/java/org/mule/runtime/internal/meta/type/DefaultTypeCatalog.java
```diff
@@ -38,25 +40,34 @@
 public final class DefaultTypeCatalog implements TypeCatalog {
 
   private List<SubTypesMappingContainer> mappings = new LinkedList<>();
-  private Map<String, ObjectType> types = new LinkedHashMap<>();
+  private Table<String, String, ObjectType> types = TreeBasedTable.create();
```
marianogonzalez: add a comment explaining what each value is

marianogonzalez: I think this is better suited by a Map<String, Pair>... I would rather move Pair to api and use here than adding guava just for this

pablomusumeci: This is more like a Map<String, Map<String, ObjectType>>

pom.xml
```diff
@@ -145,6 +146,11 @@
             <artifactId>gson</artifactId>
             <version>${gsonVersion}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
alegmarra: have you built the distros/updated the assembly?

src/main/java/org/mule/runtime/api/meta/type/TypeCatalog.java
```diff
@@ -93,4 +93,14 @@ static TypeCatalog getDefault(Set<ExtensionModel> extensions) {
...
 
+  /**
+   * @return a {@link Collection} with all the types declared by the extension with name {@code extensionName}
+   */
+  Collection<ObjectType> getExtensionTypes(String extensionName);
+
+  /**
+   * @return an {@link Optional} with the extension name that contributed with this given type
+   */
+  Optional<String> getExtension(String typeId);
```
alegmarra: getDeclaringExtension

src/main/java/org/mule/runtime/internal/meta/type/DefaultTypeCatalog.java
```diff
@@ -38,25 +40,34 @@
...
       mappings.add(new SubTypesMappingContainer(e.getSubTypes()));
-      e.getTypes().forEach(t -> getTypeId(t).ifPresent(id -> types.put(id, t)));
+      e.getTypes().forEach(t -> getTypeId(t).ifPresent(id -> types.put(e.getName(), id, t)));
     });
   }
 
   @Override
   public Optional<ObjectType> getType(String typeId) {
-    return ofNullable(types.get(typeId));
+    return types.column(typeId).values().stream().findFirst();
```
alegmarra: starting to wonder if is not better just to have two indexed structures, <typeId, type>; <typeId, extension> (not extensionName, the actual extension)

marianogonzalez: the answer to that depends on how and how often is this structure queried

alegmarra: lots

PR #232, opened 2017-05-30
MULE-12645: use standard license header mechanism from mule

alepulver:


pom.xml
```diff
@@ -231,9 +232,46 @@
...
+                            <exclude>**/*.xml</exclude>
+                            <exclude>**/*.properties</exclude>
+                            <exclude>**/*.sample</exclude>
+                            <exclude>**/*.md</exclude>
+                            <exclude>**/*.xsl</exclude>
+                            <exclude>**/*.html</exclude>
+                            <exclude>**/*.css</exclude>
+                            <exclude>**/build-number.txt</exclude>
+
+                            <exclude>**/org/springframework/**</exclude>
```
anosenzo: All the excludes of particular packages or classes should be removed because are not needed.

PR #237, opened 2017-06-01
MULE-12634: Add a getNamedParts() method in MultiPartPayload

elrodro83:


src/main/java/org/mule/runtime/api/message/MultiPartPayload.java
```diff
@@ -7,11 +7,12 @@
...
 
 import java.util.List;
+import java.util.Map;
 import java.util.NoSuchElementException;
 
 /**
- * Represents a payload of a {@link Message} composed of many different parts. Each parts is in itself a
- * {@link Message}, and has {@code attributes} specific to that parts (such as the headers of a single http part).
+ * Represents a payload of a {@link Message} composed of many different parts. Each parts is in itself a {@link Message}, and has
+ * {@code attributes} specific to that parts (such as the headers of a single http part).
```
afelisatti: I'm having a Kraan inspired moment but `those parts` and `HTTP`.

src/main/java/org/mule/runtime/api/message/MultiPartPayload.java
```diff
@@ -36,4 +37,9 @@
    */
   Message getPart(String partName);
 
+  /**
+   * @return the contained parts by their ginven name.
```
afelisatti: `given`

PR #239, opened 2017-06-02
MULE-12664 Fix SOAP Connect message builder to match WSC builder

alegmarra:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/HasDisplayModelDeclarer.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public interface HasDisplayModelDeclarer<T> {
+
+  /**
+   * Sets the given {@code displayModel}
+   *
+   * @param displayModel a {@link DisplayModel}
+   * @return {@code this} declarer
+   */
+  T withDisplay(DisplayModel displayModel);
```
marianogonzalez: withDisplayModel, so that it's consistent with HasDisplayModel

PR #240, opened 2017-06-08
MULE-12243: JMS should publish messages with String Representable Mim…

estebanwasinger:
…eTypes as StringMessages

### DIFF
--- a/src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
+++ b/src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
@@ -7,9 +7,9 @@
 package org.mule.runtime.api.metadata;
 
 import static java.util.Collections.unmodifiableList;
-import static java.util.stream.Collectors.toList;
 
 import java.util.List;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
@@ -21,20 +21,20 @@ import java.util.stream.Stream;
 public final class MediaTypeUtils {
 
   private static final String TEXT = "text";
-  private static final List<MediaType> STRING_REPRESENTABLE_MIME_TYPES = Stream.of(
-                                                                                   "application/json",
-                                                                                   "application/xml",
-                                                                                   "application/javascript",
-                                                                                   "application/rtf",
-                                                                                   "application/csv",
-                                                                                   "application/vnd.mozilla.xul+xml",
-                                                                                   "application/x-csh",
-                                                                                   "application/x-sh",
-                                                                                   "application/xhtml+xml",
-                                                                                   "image/svg+xml",
-                                                                                   "text/*")
+  private static final List<MediaType> STRING_REPRESENTABLE_MIME_TYPES = unmodifiableList(Stream.of(
+                                                                                                    "application/json",
+                                                                                                    "application/xml",
+                                                                                                    "application/javascript",
+                                                                                                    "application/rtf",
+                                                                                                    "application/csv",
+                                                                                                    "application/vnd.mozilla.xul+xml",
+                                                                                                    "application/x-csh",
+                                                                                                    "application/x-sh",
+                                                                                                    "application/xhtml+xml",
+                                                                                                    "image/svg+xml",
+                                                                                                    "text/*")
       .map(MediaType::parse)
-      .collect(toList());
+      .collect(Collectors.toList()));
 
 
   private MediaTypeUtils() {}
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,44 @@
...
+                                                                                  "application/xhtml+xml",
+                                                                                  "image/svg+xml",
+                                                                                  "text/calendar",
+                                                                                  "text/css",
+                                                                                  "text/csv",
+                                                                                  "text/json",
+                                                                                  "text/html",
+                                                                                  "text/plain")
+      .map(MediaType::parse)
+      .collect(toList());
```
marianogonzalez: unmodifiable

src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Utility class to share logic about {@link MediaType media types}
+ *
+ * @since 1.0
+ * @see MediaType
+ */
+public final class MediaTypeUtils {
+
+  public static final List<MediaType> STRING_REPRESENTABLE_MIME_TYPES = Stream.of(
```
marianogonzalez: should be private. Add a getStringRepresentableMimeTypes() method instead

src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+      .collect(toList());
+
+
+  private MediaTypeUtils() {}
+
+  /**
+   * @return the {@link List} of {@link MediaType} that are considered as {@link String} representable
+   */
+  public static List<MediaType> getStringRepresentableMimeTypes() {
+    return unmodifiableList(STRING_REPRESENTABLE_MIME_TYPES);
```
marianogonzalez: don't create a new unmodifiableList every time... make the static one unmodifiable and return always the same.... PUede alguien pensar en el GC!!!!

src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * Utility method which given a {@link MediaType} will verify if this one is {@link String} representable
```
marianogonzalez: Verifies that the given {@code mediaType} is representable as a String

src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * Verifies that the given {@code mediaType} is representable as a {@link String}
+   *
+   * @param mediaType to verify
+   * @return boolean indicating whether the {@link MediaType} is {@link String} representable or not
+   */
+  private static boolean isStringRepresentable(MediaType mediaType) {
+    return mediaType.getPrimaryType().equals(TEXT) || STRING_REPRESENTABLE_MIME_TYPES
```
marianogonzalez: check for null and add test cases

src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * Verifies that the given {@code mediaType} is representable as a {@link String}
+   *
+   * @param mediaType to verify
+   * @return boolean indicating whether the {@link MediaType} is {@link String} representable or not
+   */
+  public static boolean isStringRepresentable(MediaType mediaType) {
+    return mediaType.getPrimaryType().equals(TEXT) || STRING_REPRESENTABLE_MIME_TYPES
```
alegmarra: checkArgument for null

src/main/java/org/mule/runtime/api/metadata/MediaTypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Utility class to share logic about {@link MediaType media types}
+ *
+ * @since 1.0
+ * @see MediaType
+ */
+public final class MediaTypeUtils {
+
+  private static final String TEXT = "text";
+  private static final List<MediaType> STRING_REPRESENTABLE_MIME_TYPES = unmodifiableList(Stream.of(
```
alegmarra: this list should be declared in the MediaType class, not in a Utils

estebanwasinger: Why should be in `MediaType` class?

alegmarra: This list is tightly coupled with the MediaType support we provide in Mule. For me this is part of the definition of a MimeType in mule's context. Since the supported mediatypes are not going to change, this will work, but don't think is the best pattern to follow.

marianogonzalez: separate they shall be

PR #249, opened 2017-06-22
MULE-11606: Support a way to have DW handle multimaps

afelisatti:


src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+    return paramsMap.hashCode();
+  }
+
+  public Map<K, ? extends List<V>> toListValuesMap() {
+    return unmodifiableMap(paramsMap);
+  }
+
+  @Override
+  public String toString() {
+    return "ParameterMap{" + Arrays.toString(paramsMap.entrySet().toArray()) + '}';
```
afelisatti: Have to change this as well.

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Implementation of a multi-map that allows the aggregation of keys and access to the aggregated list or a single value (the
+ * last).
+ *
+ * @param <K> the type of keys maintained by this map
+ * @param <V> the type of mapped values
+ */
```
elrodro83: add `@since`

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+  private static final long serialVersionUID = -4723226524524565104L;
+
+  protected Map<K, LinkedList<V>> paramsMap;
+
+  public MultiMap(final Map<K, V> parametersMap) {
+    this.paramsMap = new HashMap<>();
+    parametersMap.forEach((key, value) -> {
+      if (value instanceof Collection) {
+        Collection<V> values = (Collection) value;
+        values.stream().forEach(collectionValue -> this.put(key, collectionValue != null ? collectionValue : null));
```
elrodro83: how about turning this around and use a map collector?

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+    Collection<V> values = paramsMap.remove(key);
+    if (values != null) {
+      return values.iterator().next();
+    }
+    return null;
+  }
+
+  @Override
+  public void putAll(Map<? extends K, ? extends V> aMap) {
+    for (K key : aMap.keySet()) {
```
elrodro83: using the entryset instead of the key and doing a lookup inside?

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+      LinkedList<V> values = new LinkedList<>();
```
elrodro83: declare as just `List`

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  @Override
+  public Set<K> keySet() {
+    return paramsMap.keySet();
+  }
+
+  @Override
+  public Collection<V> values() {
+    ArrayList<V> values = new ArrayList<>();
```
elrodro83: declare as just `List`

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+   * @return a list of values to which the specified key is mapped, or
+   *         an empty list if this map contains no mappings for the key
+   */
+  public List<V> getAll(Object key) {
+    return paramsMap.containsKey(key) ? unmodifiableList(paramsMap.get(key)) : emptyList();
+  }
+
+  @Override
+  public V put(K key, V value) {
+    LinkedList<V> previousValue = paramsMap.get(key);
```
elrodro83: declare as just `List`

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+    LinkedList<V> newValue = previousValue;
```
elrodro83: declare as just `List`

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+
+  @Override
+  public Set<K> keySet() {
+    return paramsMap.keySet();
+  }
+
+  @Override
+  public Collection<V> values() {
+    ArrayList<V> values = new ArrayList<>();
+    for (K key : paramsMap.keySet()) {
```
elrodro83: use the entryset

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+      values.add(paramsMap.get(key).getLast());
+    }
+    return values;
+  }
+
+  @Override
+  public Set<Entry<K, V>> entrySet() {
+    HashSet<Entry<K, V>> entries = new HashSet<>();
```
elrodro83: declare as just `Set`

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+    for (K key : paramsMap.keySet()) {
```
elrodro83: use the entrySet

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Returns a {@link List} view of the mappings contained in this map, including an entry for each value associated to the same
+   * key.
+   *
+   * @return a list view of the mappings contained in this map
+   */
+  public List<Entry<K, V>> entryList() {
+    List<Entry<K, V>> entries = new LinkedList<>();
+    for (K key : paramsMap.keySet()) {
```
elrodro83: use the entryset

src/main/java/org/mule/runtime/api/util/MultiMap.java
afelisatti: Most LinkedList references cannot be "abstracted" because the inner map values are LinkedList so as to have access to the last value for regular get, entry set and so on.

src/main/java/org/mule/runtime/api/util/MultiMap.java
afelisatti: Can't because of the last access part.

src/main/java/org/mule/runtime/api/util/MultiMap.java
afelisatti: Same as above.

src/main/java/org/mule/runtime/api/util/MultiMap.java
afelisatti: Not sure what you meant here.

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,208 @@
...
+      previousValue = new LinkedList<>(previousValue);
+    } else {
+      newValue = new LinkedList<>();
+    }
+    newValue.add(value);
+    paramsMap.put(key, newValue);
+    if (previousValue == null || previousValue.isEmpty()) {
+      return null;
+    }
+    return previousValue.getFirst();
```
elrodro83: shouldn't this be `getLast()`?

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,208 @@
...
+    return paramsMap.isEmpty();
+  }
+
+  @Override
+  public boolean containsKey(Object key) {
+    return paramsMap.containsKey(key);
+  }
+
+  @Override
+  public boolean containsValue(Object value) {
```
machaval: Shouldn't this check in the elements of the list. If not this is only going to work if value is the list of all the elements

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -0,0 +1,208 @@
...
+  public MultiMap() {
+    this.paramsMap = new LinkedHashMap<>();
+  }
+
+  public MultiMap<K, V> toImmutableMultiMap() {
+    return new MultiMap(this.paramsMap);
+  }
+
+  @Override
+  public int size() {
```
machaval: According to java map spec this is 
```Returns the number of key-value mappings in this map. If the map contains more than Integer.MAX_VALUE elements, returns Integer.MAX_VALUE.```
In my opinion we should return the size of all the elements

File: src/main/java/org/mule/runtime/api/util/MultiMap.java [hunk 16]
```diff
@@ -0,0 +1,208 @@
...
+   */
+  public List<Entry<K, V>> entryList() {
+    List<Entry<K, V>> entries = new LinkedList<>();
+    paramsMap.forEach((key, values) -> values.forEach(value -> entries.add(new AbstractMap.SimpleEntry<>(key, value))));
+    return entries;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    return paramsMap.equals(o);
```
machaval: This is not going to work this should check that it is a MultiMap and then compare with the paramsMap

src/main/java/org/mule/runtime/api/util/MultiMap.java
afelisatti: I've change it to honour that.

src/main/java/org/mule/runtime/api/util/MultiMap.java
afelisatti: Done.

PR #251, opened 2017-06-22
MULE-12889: Metadata is not resolved for SOAP-based connectors

alegmarra:
(cherry picked from commit cf23ab2)

### DIFF
--- a/src/main/java/org/mule/runtime/api/metadata/MetadataKeyBuilder.java
+++ b/src/main/java/org/mule/runtime/api/metadata/MetadataKeyBuilder.java
@@ -6,12 +6,12 @@
  */
 package org.mule.runtime.api.metadata;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/metadata/MetadataKeyBuilder.java
```diff
@@ -75,7 +74,29 @@ public MetadataKeyBuilder withProperty(MetadataProperty property) {
    * @return {@code this} builder with a new child.
    */
   public MetadataKeyBuilder withChild(MetadataKeyBuilder metadataKeyBuilder) {
-    childs.add(metadataKeyBuilder);
+    childs.add(metadataKeyBuilder.build());
```
estebanwasinger: this should not be built directly here, these builders should be built when the whole key is being built

PR #253, opened 2017-06-26
MULE-12856: Implement Parameters Options Resolver in SDK's Side

estebanwasinger:


src/main/java/org/mule/runtime/api/metadata/MetadataKey.java
```diff
@@ -7,6 +7,7 @@
 package org.mule.runtime.api.metadata;
 
 import org.mule.metadata.api.model.MetadataType;
+import org.mule.runtime.api.options.Option;
```
marianogonzalez: unused?

src/main/java/org/mule/runtime/api/options/OptionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+
+public class OptionBuilder {
+
+  private final String id;
+  private String displayName;
+  private Set<OptionBuilder> childs = new LinkedHashSet<>();
```
marianogonzalez: final

src/main/java/org/mule/runtime/api/options/OptionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+  private String partName = "";
+
+  protected OptionBuilder(String id) {
```
marianogonzalez: private

src/main/java/org/mule/runtime/api/options/OptionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+    this.id = id;
+  }
+
+  protected OptionBuilder(String id, String partName) {
```
marianogonzalez: private

src/main/java/org/mule/runtime/api/options/DefaultOption.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals;
+import static org.apache.commons.lang3.builder.HashCodeBuilder.reflectionHashCode;
+import static org.apache.commons.lang3.builder.ToStringBuilder.reflectionToString;
+import static org.apache.commons.lang3.builder.ToStringStyle.JSON_STYLE;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+/**
+ * Created by estebanwasinger on 6/19/17.
+ */
+public class DefaultOption implements Option {
```
alegmarra: we need to come up with a better name for this

src/main/java/org/mule/runtime/api/options/Option.java
```diff
@@ -0,0 +1,39 @@
...
+import org.mule.runtime.api.metadata.MetadataKey;
+
+import java.util.Set;
+
+/**
+ * mule-runtime-project
+ *
+ * @author Esteban Wasinger (http://github.com/estebanwasinger)
+ */
+public interface Option {
```
alegmarra: `Option` applies to a million things, I get the idea but is way too generic

src/main/java/org/mule/runtime/api/options/OptionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+  public static OptionBuilder newOption(String id, String partName) {
+    return new OptionBuilder(id, partName);
+  }
+
+  public OptionBuilder withDisplayName(String displayName) {
+    this.displayName = displayName;
+    return this;
+  }
+
+  public OptionBuilder withChild(OptionBuilder metadataKeyBuilder) {
```
alegmarra: is better to receive an already built child instead of receiving the builder. I know that is how the metadata keys are now, those are wrong too. Is better for the api to have: 
``` newOption(stiv).withDisplayName(Stib).withChild(newOption(bazinga).withDisplayName(Wasinga).build()).build()```
than having: 
``` newOption(stiv).withDisplayName(Stib).withChild(newOption(bazinga).withDisplayName(Wasinga)).build()```

Since in the first case you are certain that the child will be built only once, is more friendly for caching and eliminates sideeffects of reusing the same option in multiple option trees.

File: src/main/java/org/mule/runtime/api/options/OptionsProvider.java [hunk 8]
```diff
@@ -0,0 +1,16 @@
...
+ */
+package org.mule.runtime.api.options;
+
+import java.util.Set;
+
+public interface OptionsProvider {
+
+  Set<Option> getOptions(String parameterName);
+
+  Set<Option> getGroupOptions(String groupName);
```
alegmarra: why? don't assume we have duplicated param/group names, if they ask for an option with a groupName then that's the group option

alegmarra: why is this API?

src/main/java/org/mule/runtime/api/values/DefaultValue.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.apache.commons.lang3.builder.ToStringStyle.JSON_STYLE;
+
+import java.util.Set;
+
+/**
+ * Default implementation of {@link Value}
+ *
+ * @since 1.0
+ */
+public class DefaultValue implements Value {
```
marianogonzalez: ImmutableValue

src/main/java/org/mule/runtime/api/meta/model/parameter/ParameterGroupModel.java
```diff
@@ -21,7 +21,8 @@
  *
  * @since 1.0
  */
-public interface ParameterGroupModel extends NamedObject, DescribedObject, HasDisplayModel, HasLayoutModel, EnrichableModel {
+public interface ParameterGroupModel
```
marianogonzalez: what happens if a group has two different sets of resolvable options? E.g.: A default group has options for Slack channels + another set of options for subscribed bots (the example might not entirely make sense but you get the point)

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
```diff
@@ -0,0 +1,95 @@
...
+ * Model for {@link ParameterModel} and {@link ParameterGroupModel} to communicate if one of these are capable of
+ * provide {@link Value values}.
+ * <p>
+ * The element with this model will considered as a one that provides values, and also this model communicates
+ * how is the structure of th
+ * @since 1.0
+ */
+public class ValuesProviderModel {
+
+  private final List<String> requiredParameters;
```
marianogonzalez: requiredParameters is a misleading name. These are more like `suscribedParameters` or `actingParameters`or something of that sort

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Model for {@link ParameterModel} and {@link ParameterGroupModel} to communicate if one of these are capable of
+ * provide {@link Value values}.
+ * <p>
+ * The element with this model will considered as a one that provides values, and also this model communicates
+ * how is the structure of th
+ * @since 1.0
+ */
+public class ValuesProviderModel {
```
marianogonzalez: I don't see how this applies to an individual ParameterModel. It does make sense for the group, but not the parameter itself. 

* How would the state of a parameter related instance be different from that of the group?
* Isn't the order the only thing that is actually related to the parameter? (If the answer is yes, you can take the order from the requiredParameters list in the group model)

src/main/java/org/mule/runtime/api/values/DefaultValue.java
```diff
@@ -0,0 +1,83 @@
...
+import static java.util.Collections.unmodifiableSet;
+import static org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals;
+import static org.apache.commons.lang3.builder.HashCodeBuilder.reflectionHashCode;
+import static org.apache.commons.lang3.builder.ToStringBuilder.reflectionToString;
+import static org.apache.commons.lang3.builder.ToStringStyle.JSON_STYLE;
+
+import java.util.Set;
+
+/**
+ * Default implementation of {@link Value}
```
marianogonzalez: Default -> Immutable

src/main/java/org/mule/runtime/api/resolving/ExtensionResolvingContext.java
```diff
@@ -0,0 +1,42 @@
...
+package org.mule.runtime.api.resolving;
+
+import org.mule.metadata.api.ClassTypeLoader;
+import org.mule.metadata.api.builder.BaseTypeBuilder;
+import org.mule.runtime.api.connection.ConnectionException;
+import org.mule.runtime.api.lifecycle.Disposable;
+
+import java.util.Optional;
+
+public interface ExtensionResolvingContext extends Disposable {
```
marianogonzalez: jdoc

src/main/java/org/mule/runtime/api/values/ComponentValuesProvider.java
```diff
@@ -0,0 +1,24 @@
...
+package org.mule.runtime.api.values;
+
+import java.util.Set;
+
+/**
+ * This interface allows the exposure of the {@link Value values} associated to a parameter of a Operation or Source.
+ *
+ * @since 1.0
+ */
+public interface ComponentValuesProvider {
```
marianogonzalez: Don't see a reason why this should be here. Seems to me this actually belongs in ext-api

src/main/java/org/mule/runtime/api/values/ConfigurationParameterValuesProvider.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Set;
+
+/**
+ * This interface allows the exposure of the {@link Set} of {@link Value values} associated to a Configuration's or
+ * Connection Provider's parameter.
+ *
+ * @since 1.0
+ */
+public interface ConfigurationParameterValuesProvider {
```
marianogonzalez: Don't see a reason why this should be here. Seems to me this actually belongs in ext-api

src/main/java/org/mule/runtime/api/values/DefaultValue.java
marianogonzalez: Don't see a reason why this should be here. Seems to me this actually belongs in ext-api

src/main/java/org/mule/runtime/api/values/Value.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Represents a possible and valid value for a parameter.
+ * <p>
+ * Also this {@link Value} can be composed by other {@link Value Options}
+ * to form a composed option.
+ *
+ * @since 1.0
+ */
+public interface Value {
```
marianogonzalez: Don't see a reason why this should be here. Seems to me this actually belongs in ext-api

src/main/java/org/mule/runtime/api/values/ValueBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Stream;
+
+/**
+ * An implementation of the builder design pattern to create a new {@link ValueBuilder} instance.
+ *
+ * @since 1.0
+ */
+public class ValueBuilder {
```
marianogonzalez: Don't see a reason why this should be here. Seems to me this actually belongs in ext-api

src/main/java/org/mule/runtime/api/values/ValueResolvingException.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.api.i18n.I18nMessage;
+
+/**
+ * {@link Exception} to indicate than an error occurred resolving {@link Value values}
+ *
+ * @since 1.0
+ */
+public class ValueResolvingException extends MuleException {
```
marianogonzalez: Don't see a reason why this should be here. Seems to me this actually belongs in ext-api

src/main/java/org/mule/runtime/api/values/ComponentValuesProvider.java
estebanwasinger: This should be here, this is the interface that any `Message Processor` should implement if they want to have the capability to provide `Values` for their parameters.
This class is the cousin of `MetadataKeyProvider`

src/main/java/org/mule/runtime/api/meta/model/parameter/ParameterGroupModel.java
estebanwasinger: if the `ValueProviderModel` is applied to a `ParameterGroup` this means that the entire `ParameterGroup` is an element that can provide `Values`.
In case that you require `ValueProvider` enabled parameters inside a `PramGroup`, each `ParameterModel` will have this `ValueProviderModel`

src/main/java/org/mule/runtime/api/values/ComponentValuesProvider.java
marianogonzalez: but unlike metadata, I don't think we're interested in non SDK components providing option resolvers

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
estebanwasinger: As said in the previous comment, if the `Value` is single level, the model property will only be applied to a `Parameter`, in case that is a "Multi Level" Value, the model property will be applied to the `ParameterGroup`.
> How would the state of a parameter related instance be different from that of the group?

I don't get it

> Isn't the order the only thing that is actually related to the parameter? (If the answer is yes, you can take the order from the requiredParameters list in the group model)

The order is customizable by the extension developer an also not all parameters inside a parameter group will be part of the multi level value.

src/main/java/org/mule/runtime/api/meta/model/parameter/ParameterGroupModel.java
marianogonzalez: but then the model becomes ambigous because there're two ways of signaling the same thing

src/main/java/org/mule/runtime/api/values/Value.java
estebanwasinger: this should be public is what is being exposed through, the next to be created, `ValueProviderService` and what the agent and tooling will consume.

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterGroupDeclaration.java
```diff
@@ -9,6 +9,7 @@
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 import org.mule.runtime.api.meta.model.display.LayoutModel;
 import org.mule.runtime.api.meta.model.parameter.ParameterGroupModel;
+import org.mule.runtime.api.meta.model.parameter.ValuesProviderModel;
```
marianogonzalez: remove

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
marianogonzalez: forgot this one

src/main/java/org/mule/runtime/api/meta/model/parameter/HasValuesProviderModel.java
```diff
@@ -0,0 +1,22 @@
...
+package org.mule.runtime.api.meta.model.parameter;
+
+import java.util.Optional;
+
+/**
+ *  A model which optionally contains a {@link ValuesProviderModel}
+ *
+ * @since 1.0
+ */
+public interface HasValuesProviderModel {
```
marianogonzalez: if only ParameterModel is going to implement this interface, not sure of its value.

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
```diff
@@ -0,0 +1,91 @@
...
+ * Model for {@link ParameterModel} and {@link ParameterGroupModel} to communicate if one of these are capable of
+ * provide {@link Value values}.
+ * <p>
+ * The element with this model will considered as a one that provides values.
+ * @since 1.0
+ */
+public class ValuesProviderModel {
+
+  private final List<String> requiredParameters;
+  private final Integer partOrder;
```
marianogonzalez: shouldn't this be Optional<Integer> for the case in which there's no multilevel?

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+
+/**
+ * Model for {@link ParameterModel} and {@link ParameterGroupModel} to communicate if one of these are capable of
+ * provide {@link Value values}.
+ * <p>
+ * The element with this model will considered as a one that provides values.
+ * @since 1.0
+ */
+public class ValuesProviderModel {
```
marianogonzalez: ValueProviderModel... Drop the 's'

src/main/java/org/mule/runtime/api/meta/model/parameter/HasValuesProviderModel.java
alegmarra: +1, we can always refactor it out if some other model needs this

src/main/java/org/mule/runtime/api/meta/model/parameter/ValuesProviderModel.java
```diff
@@ -0,0 +1,91 @@
...
+
+import static org.mule.runtime.api.util.Preconditions.checkNotNull;
+import org.mule.runtime.api.values.Value;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import java.util.List;
+
+/**
+ * Model for {@link ParameterModel} and {@link ParameterGroupModel} to communicate if one of these are capable of
```
alegmarra: not used in ParameterGroupModel

PR #259, opened 2017-06-30
MULE-9249: Create @Path to allow users to use file/directory pickers …

juandesi:
…as editors for their mule extensions

### DIFF
--- a/src/main/java/org/mule/runtime/api/meta/model/display/DisplayModel.java
+++ b/src/main/java/org/mule/runtime/api/meta/model/display/DisplayModel.java
@@ -7,6 +7,7 @@
 package org.mule.runtime.api.meta.model.display;
 
 import java.util.Objects;
+import java.util.Optional;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/meta/model/display/DisplayModel.java
```diff
@@ -60,6 +61,17 @@ public DisplayModelBuilder example(String example) {
     }
 
     /**
+     * A {@link PathModel} with metadata for a parameter that points to file or directory.
+     *
+     * @param pathModel a {@link PathModel} with the metadata associated to the path param.
+     * @return {@code this} builder
+     */
+    public DisplayModelBuilder path(PathModel pathModel) {
```
marianogonzalez: Shouldn't this go on the LayoutModel? It feels like it belongs more with things like `password` or `text` than with `summary` and `example`

PR #271, opened 2017-07-07
MULE-12857: Implement Parameters Options Resolver in Mule's Side

estebanwasinger:


src/main/java/org/mule/runtime/api/values/ImmutableValueResult.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.values;
```
marianogonzalez: per convention, package names need to be singular

PR #283, opened 2017-07-20
MULE-13003: Review mutability of Authentication object

elrodro83:


src/main/java/org/mule/runtime/api/security/Authentication.java
```diff
@@ -10,24 +10,16 @@
...
 
   /**
-   * @param authenticated {@code true} if {@code this} {@link Authentication} has been authenticated
-   */
-  void setAuthenticated(boolean authenticated);
-
-  /**
-   * @return {@code true} if {@code this} {@link Authentication} has been authenticated
-   */
-  boolean isAuthenticated();
```
afelisatti: Why was this necessary?

elrodro83: don't know, it was only used in assetions in tests, no usage in productive code.

PR #284, opened 2017-07-21
MULE-13160: CleanUp ObjectStore API

marianogonzalez:


src/main/resources/META-INF/mule-module.properties
```diff
@@ -43,6 +43,7 @@ artifact.export.classPackages=org.mule.runtime.api.app.declaration,\
                               org.mule.runtime.api.metadata,\
                               org.mule.runtime.api.metadata.descriptor,\
                               org.mule.runtime.api.metadata.resolving,\
+                              org.mule.runtime.api.deployment.persistence,\
```
pablolagreca: What's the purpose of exporting this package here if there was no classes added to that package?

src/main/java/org/mule/runtime/api/store/ObjectStoreManager.java
```diff
@@ -7,76 +7,65 @@
...
+ * Any component in need to use an {@link ObjectStore} should created it through an implementation of this interface.
+ * <p>
+ * All {@link ObjectStore} instances created through an instance of this interface, should also be destroyed through the
+ * same instance.
+ * <p>
+ * Implementations are required to be thread-safe.
  *
  * @since 1.0
  */
 public interface ObjectStoreManager {
```
pablolagreca: Add to this interface the keys for the default transient and persistent base object stores so they can be referenced by custom implementation to override them.

src/main/resources/META-INF/mule-module.properties
marianogonzalez: i was getting a build error because of this missing. It's probably a left over from a rebase.

src/main/java/org/mule/runtime/api/store/ObjectStoreManager.java
marianogonzalez: the problem with doing that is that it assumes there're base object stores. I believe that to be an implementation detail of the MuleObjectStoreManager. A RedisObjectStore manager might not need that concept

src/main/java/org/mule/runtime/api/store/ObjectStoreManager.java
```diff
@@ -6,77 +6,87 @@
...
- *
- * Any {@link ObjectStore} required by any component should be obtained through this API
+ * Creates and manages {@link ObjectStore} instances.
+ * <p>
+ * Any component in need to use an {@link ObjectStore} should created it through an implementation of this interface.
+ * <p>
+ * This manager works under the concept of &quot;Base Object Stores&quot;. This means that all created Object Stores
+ * are actually partitions on these base stores. This manager is in charge of creating and handling such partitions
+ * and exposing them as stand alone stores. There're two base object stores, one for the in memory ones and another for
+ * the transient ones.
```
GeraldLoeffler: @marianogonzalez : should this read "persistent" instead of "transient"?

PR #287, opened 2017-07-25
MULE-11675 - Support DI for mule objects into spring beans

pablolagreca:


src/main/java/org/mule/runtime/api/ioc/ConfigurableObjectProvider.java
```diff
@@ -0,0 +1,32 @@
...
+ * 
+ * @since 1.0
+ */
+public interface ConfigurableObjectProvider extends ObjectProvider {
+
+  /**
+   * Method to be called to prepare the {@link ConfigurableObjectProvider}. It is expected that the provider is ready to be used
+   * after calling this method.
+   * <p/>
+   * This method will be invoke before {@link Initialisable#initialise()} if the implementation implements that interface.
```
elrodro83: invoke -> invoked

src/main/java/org/mule/runtime/api/ioc/ObjectProvider.java
```diff
@@ -15,23 +14,12 @@
 /**
  * Interface meant to be implemented by components that will provide objects that may be referenced from mule configuration files.
  * <p/>
```
elrodro83: remove this line

PR #289, opened 2017-07-25
MULE-12872 Add payload length to TypedValue API

dfeist:


src/main/java/org/mule/runtime/api/metadata/TypedValue.java
```diff
@@ -66,12 +70,32 @@
...
       this.dataType = fromObject(value);
     } else {
       this.dataType = dataType;
     }
+    if (length.isPresent()) {
+      this.length = length.get();
+    } else if (value instanceof byte[]) {
+      this.length = ((byte[]) value).length;
+    } else if (value instanceof String) {
+      this.length = ((String) value).getBytes(this.dataType.getMediaType().getCharset().orElse(defaultCharset())).length;
```
elrodro83: shouldn't this be the length of the string? If not, just length as the attribute name is misleading, since it's actually the size of the byte array representation of the payload

src/main/java/org/mule/runtime/api/metadata/TypedValue.java
```diff
@@ -90,4 +114,13 @@ public T getValue() {
     return value;
   }
 
+  /**
+   * If available obtain the length (in bytes) of the valye.
+   * 
+   * @return length of the value in bytes.
+   */
+  public Optional<Long> getLength() {
+    return length >= 0 ? Optional.of(length) : empty();
```
elrodro83: this is inconsistent with setting `-1`a couple of lines above. Isn't 0 a valid value?

src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -90,19 +90,20 @@ static Message of(Object payload) {
...
+     * This will replace the whole payload including both the value and {@link DataType}.
      *
-     * @param mediaType the mediaType to set
-     * @return this builder
-     * @throws NullPointerException if the mediaType is null
+     * @param payload the message payload
+     * @return this builder.
      */
-    Builder mediaType(MediaType mediaType);
+    Builder typedPayload(TypedValue payload);
```
elrodro83: I would prefer adding a `length` method rather than allowing to path the wholy build TypedValue. If this is done like this, we would be losing the "wizard-like builder" we intended to implement at first. Maybe this method makes sense to have in the implementation to use internally, but i wouldn't expose it in the API.

src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -162,28 +172,36 @@ static Message of(Object payload) {
...
     /**
+     * Sets the attributes given {@code value} using a {@link TypedValue}.
+     * <p>
+     * This will replace the whole payload including both the value and {@link DataType}.
+     *
+     * @param attributes the attributes TypedValue
+     * @return this builder
+     * @throws NullPointerException if the value is null
+     */
+    Builder typedAttributes(TypedValue attributes);
```
elrodro83: same as comment above

src/main/java/org/mule/runtime/api/metadata/TypedValue.java
dfeist: The javadoc clearly states that it's the "length in bytes", i can add detail to javadoc to make it clear it takes into account the charset in the case of a String. Will this suffice, if not I accept suggestion for something more descriptive than `length`..

src/main/java/org/mule/runtime/api/metadata/TypedValue.java
dfeist: Yes `0` is valid, that is why this is `>=0` and not `>0`. I'm not sure what the issue is here.  I'm using  -1 instead of an Optional field as an Optional field is bad practice and not serializable. Could use nullable Integer, but prefer not to.

elrodro83: oops, i misread it

src/main/java/org/mule/runtime/api/metadata/TypedValue.java
elrodro83: Ok, just explaining this particular case in the javadoc would do.

PR #297, opened 2017-07-31
MULE-13032: Review all interfaces builders to match DW standard

afelisatti:


afelisatti:
REVIEW ONLY

src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -70,17 +70,27 @@ static Builder builder(Message message) {
...
    */
   static Message of(Object payload) {
-    return builder().payload(payload).build();
+    return builder().value(payload).build();
   }
 
   interface PayloadBuilder {
 
     /**
+     * Sets the {@link Message}'s payload to be built.
```
dfeist: Improve javadoc to be more explicit about the difference between value() and payload() and when to use which.

src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -93,7 +103,7 @@ static Message of(Object payload) {
      * @param value the message content
```
dfeist: Improve javadoc to be more explicit about the difference between value() and payload() and when to use which.

src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -70,17 +70,36 @@ static Builder builder(Message message) {
...
+   */
   interface PayloadBuilder {
 
     /**
+     * Sets the {@link Message}'s {@link TypedValue} payload to be built, including it's value and media type. This should be used
+     * when you already have a {@link TypedValue}. Alternatively, a {@link Message} can be constructed from just the value or value
+     * and mediaType component parts and the builder will generate the {@link TypedValue}.
+     * See {@link #value(Object)}, {@link #mediaType(MediaType)}.
+     * <p>
+     * If a {@link DataType} has previously been set it will be updated to reflect the type of the new payload.
```
dfeist: `If a {@link DataType} has previously it will be updated to reflect the type of the new payload.` -> `If a MediaType or Value have been set then they will be overwritten`

also, put @see annotation below, not in tex.

PR #299, opened 2017-08-07
MULE-11553 - Create an API to create beans that behave like flows

pablolagreca:


src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,233 @@
...
+ * <p/>
+ * Instances of this interface can be build using {@link Event#builder()} static method.
+ * <p/>
+ * 
+ * @since 1.0
+ */
+public interface Event {
+
+  /**
+   * Returns the variables in the event
```
afelisatti: `current event`

afelisatti: or just `event` throughout

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,233 @@
...
+   * The returned value will depend on the source that created this event, and the flow that is executing the event.
+   *
+   * @return the correlation id to use for this event.
+   */
+  String getCorrelationId();
+
+  /**
+   * Creates a {@link BindingContext} from the event.
+   * 
+   * @return the event context.SourceWithFlowInfoTestCase
```
afelisatti: this reference to a test case looks odd...

src/main/java/org/mule/runtime/api/security/SecurityContext.java
```diff
@@ -0,0 +1,27 @@
...
+ * {@code SecurityContext} holds security information and is associated with the MuleSession.
+ * 
+ * @since 4.0
+ */
+public interface SecurityContext extends Serializable {
+
+  /**
+   * @param authentication the {@link Authentication} to set in the current context
+   */
+  void setAuthentication(Authentication authentication);
```
afelisatti: This should be read only at this level.

src/main/java/org/mule/runtime/api/security/SecurityContext.java
```diff
@@ -0,0 +1,27 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.security;
+
+import java.io.Serializable;
+
+/**
+ * {@code SecurityContext} holds security information and is associated with the MuleSession.
+ * 
+ * @since 4.0
```
afelisatti: 1.0

src/main/java/org/mule/runtime/internal/el/BindingContextUtils.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mule.runtime.api.metadata.TypedValue;
+import org.mule.runtime.api.security.Authentication;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Provides a reusable way for creating {@link BindingContext}s.
+ * 
+ * @since 4.0
```
afelisatti: 1.0

src/main/java/org/mule/runtime/internal/el/BindingContextUtils.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+public class BindingContextUtils {
+
+  public static final String PAYLOAD = "payload";
+  public static final String DATA_TYPE = "dataType";
+  public static final String ATTRIBUTES = "attributes";
+  public static final String ERROR = "error";
+  public static final String CORRELATION_ID = "correlationId";
+  public static final String VARIABLES = "variables";
```
afelisatti: vars

src/main/java/org/mule/runtime/internal/el/BindingContextUtils.java
```diff
@@ -0,0 +1,96 @@
...
+  public static BindingContext addEventBindings(Event event, BindingContext baseContext) {
+    requireNonNull(event);
+    requireNonNull(baseContext);
+
+    BindingContext.Builder contextBuilder = BindingContext.builder(baseContext);
+
+    Map<String, TypedValue> flowVars = new HashMap<>();
+    event.getVariables().forEach((name, value) -> {
+      flowVars.put(name, value);
+      contextBuilder.addBinding(name, value);
```
afelisatti: This is really behind master. Several clean up changes were done in addition to this.

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Default implementation for {@link Event.Builder}.
+ * 
+ * @since 1.0
+ */
+public class EventBuilderImplementation implements Event.Builder {
```
afelisatti: Should we expose this here? We are not doing it for Message nor BindingContext AFAIK. We are using SPI to load the implementations from core.

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
```diff
@@ -0,0 +1,129 @@
...
+
+  private String correlationId;
+  private Message message;
+  private Map<String, TypedValue<?>> variables = new HashMap<>();
+  private Map<String, TypedValue<?>> parameters = new HashMap<>();;
```
afelisatti: `;;`

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
```diff
@@ -0,0 +1,129 @@
...
+  private Map<String, TypedValue<?>> properties = new HashMap<>();;
```
afelisatti: Same

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * Add a variable.
+     *
+     * @param key the key of the variable to add.
+     * @param value the value of the variable to add. {@code null} values are supported.
+     * @param mediaType additional metadata about the {@code value} type.
+     * @return the builder instance
+     */
+    Builder addVariable(String key, Object value, DataType mediaType);
```
afelisatti: `mediaType` -> `dataType`

src/main/java/org/mule/runtime/internal/event/EventImplementation.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Internal implementation of {@link Event}
+ * 
+ * @since 1.0
+ */
+public class EventImplementation implements Event {
```
afelisatti: Same as the builder. Should we expose the implementation in mule-api?

src/main/java/org/mule/runtime/internal/event/EventImplementation.java
```diff
@@ -0,0 +1,104 @@
...
+  private String correlationId;
+  private Message message;
+  private Map<String, TypedValue<?>> variables;
+  private Map<String, TypedValue<?>> parameters;
+  private Map<String, TypedValue<?>> properties;
+  private Error error;
+  private EventContext eventContext;
+  private SecurityContext securityContext;
+
+  public EventImplementation(Message message, Map<String, TypedValue<?>> variables,
```
afelisatti: Shouldn't this be at least protected?

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
dfeist: The `Implementation` suffix we have being using for single/internal private/protected implementations inside builders only. eg. `MessageImplementation` (in DefaultMessageBuilder). This isn't the case here, and this should follow normal naming standard i.e. `SomethingDescriptiveEventBuilder`.  This needs to be consistent with `Message/MessageBuilder` too (both naming and if it exists in mule-api or not)

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,233 @@
...
+   */
+  interface Builder {
+
+    /**
+     * Sets the configuration provided by the {@link org.mule.runtime.api.event.Event} into the builder
+     *
+     * @param event the event to get the data from
+     * @return the builder instance
+     */
+    Builder from(org.mule.runtime.api.event.Event event);
```
dfeist: Use `static Builder builder(Event event)` in `Event` instead of this, to be consistent with existing builders.

(or we need to change other builder API's to be consistent)

src/main/java/org/mule/runtime/internal/event/EventImplementation.java
dfeist: With other immutable objects, we use private/protected class defined inside builder itself, not outside of builder and for sure not public.

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,233 @@
...
+
+  /**
+   * Creates a {@link BindingContext} from the event.
+   * 
+   * @return the event context.SourceWithFlowInfoTestCase
+   */
+  BindingContext asBindingContext();
+
+  static Builder builder() {
+    return new EventBuilderImplementation();
```
dfeist: You should not be able to create an `Event` without an i) existing event ii) `EventContext`.  This permits creation of an event without either.

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
pablolagreca: We discuss that with PK and that have disadvantages that we have to solve. For instance, how would I use the builder if I only reference the API jar and the core is not in the classpath

src/main/java/org/mule/runtime/api/event/Event.java
pablolagreca: Not sure I understood. There's no concept of "current" event.

src/main/java/org/mule/runtime/api/event/Event.java
pablolagreca: As we discuss, it can be allowed since this Event is not the one that can is used within flows

dfeist: Understand this now yes. *But* the whole two events thing needs some more though IMO. In fact, I'd re-think this and introduce it if required as part of flow execution PR given we'll likely need to define a service interface and/or template for doing this and we may be able to replace the need to create an event for this use case with simply populating a template with message and vars, and this allow us to avoid the whole two event thing.

_(we have 2 messages interfaces which is not ideal, but it's different because i) it's required for compatibility module ii) we always use the same implementation, not two)_

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,238 @@
...
+   * @return the context for this session or null if the request is not secure.
+   */
+  SecurityContext getSecurityContext();
+
+  /**
+   * The returned value will depend on the source that created this event, and the flow that is executing the event.
+   *
+   * @return the correlation id to use for this event.
+   */
+  String getCorrelationId();
```
dfeist: This is only in event in core for legacy compatibility reasons. The correct way to get correlationId is `event.content.correlationId`, consider removing this method unless really needed, as it means there are two ways to get correlationId.

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,238 @@
...
+   */
+  EventContext getContext();
+
+  /**
+   * The security context for this session. If not null outbound, inbound and/or method invocations will be authenticated using
+   * this context
+   *
+   * @return the context for this session or null if the request is not secure.
+   */
+  SecurityContext getSecurityContext();
```
dfeist: Consider using `Authentication` directly, `SecurityContext` seems completely redundant.  Also consider if this can/should be part of context rather than event (e.g if it is fixed for invocation of flow(s))

src/main/java/org/mule/runtime/api/event/EventContext.java
```diff
@@ -0,0 +1,52 @@
...
+   * @return the location where this context's events come from
+   */
+  ComponentLocation getOriginatingLocation();
+
+  /**
+   * Returns {@code this} context's parent if it has one
+   *
+   * @return {@code this} context's parent or {@link Optional#empty()} if it doesn't have one
+   */
+  Optional<EventContext> getParentContext();
```
dfeist: `Optional<EventContext>` -> `Optional<? extends EventContext>` to avoid `getInternalEventContext()` in core.

src/main/java/org/mule/runtime/api/security/SecurityContext.java
```diff
@@ -0,0 +1,22 @@
...
+package org.mule.runtime.api.security;
+
+import java.io.Serializable;
+
+/**
+ * {@code SecurityContext} holds security information and is associated with the MuleSession.
+ * 
+ * @since 1.0
+ */
+public interface SecurityContext extends Serializable {
```
dfeist: I think this is redundant. Use `Authentication` directly.

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
```diff
@@ -0,0 +1,214 @@
+/*
```
dfeist: `EventBuilderImplementation` -> `DefaultEventBuilder`

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,238 @@
...
+   * To avoid losing the error field after the error handler the user can define a variable pointing to the error field.
+   *
+   * @return an optional of the error associated with the event. Will be empty if there's no error associated with the event.
+   */
+  Optional<Error> getError();
+
+  /**
+   * @return the context applicable to all events created from the same root {@link Event} from a source.
+   */
+  EventContext getContext();
```
dfeist: Document that this, while not `Optional`, this will return null for `external` Event instances created via mule-api builder?

src/main/java/org/mule/runtime/api/event/Event.java
afelisatti: Some of the javadocs reference the current event while others just the event. I think it should be event throughout all of them.

src/main/java/org/mule/runtime/internal/event/EventBuilderImplementation.java
```diff
@@ -0,0 +1,214 @@
...
+
+  public EventBuilderImplementation() {}
+
+  public EventBuilderImplementation(Event event) {
+    this.message = event.getMessage();
+    this.variables = event.getVariables();
+    this.parameters = event.getParameters();
+    this.properties = event.getProperties();
+    this.error = event.getError().orElse(null);
+    this.eventContext = event.getContext();
```
afelisatti: How can the eventContext and securityContext/authentication set?

afelisatti: When not build from an existing event, I mean.

src/main/java/org/mule/runtime/api/event/Event.java
pablolagreca: I tried the approach of having a NullEventContext but adds more noise to the API. I think that if we just properly document that this method may return null when instances were created from the builder will make the aPI more clear.

src/main/java/org/mule/runtime/api/interception/InterceptionEvent.java
```diff
@@ -22,44 +23,7 @@
  *
  * @since 1.0
  */
```
dfeist: InterceptionEvent has it's own methods for setting message/variables, is there any way it could implement `InputEvent` and use a generic return type to avoid event mutation in two places in API?

src/main/resources/META-INF/mule-module.properties
```diff
@@ -60,3 +61,7 @@ artifact.export.classPackages=org.mule.runtime.api.app.declaration,\
                               org.mule.runtime.api.tx,\
                               org.mule.runtime.api.util,\
                               javax.inject
+
+artifact.privileged.classPackages=org.mule.runtime.internal.event
```
dfeist: Is this still needed? I don't think so...

src/main/java/org/mule/runtime/api/event/InputEvent.java
```diff
@@ -0,0 +1,177 @@
...
+   */
+  static InputEvent newInputEvent() {
+    return new DefaultInputEvent();
+  }
+
+  /**
+   * @param event an {@link Event} to use as source of data
+   * @return a new {@link InputEvent} with the content of an existent {@link Event}
+   */
+  static InputEvent newInputEvent(Event event) {
```
dfeist: Redundancy in  `InputEvent.newInputEvent()`, how about simply `InputEvent.ceate()` or something like that?

src/main/java/org/mule/runtime/api/event/InputEvent.java
```diff
@@ -0,0 +1,177 @@
...
+import org.mule.runtime.api.message.Message;
+import org.mule.runtime.api.metadata.DataType;
+import org.mule.runtime.api.metadata.TypedValue;
+import org.mule.runtime.internal.event.DefaultInputEvent;
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Interface for defining the content of an event.
```
dfeist: Would be useful to day when/where InputEvent should be used..

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -0,0 +1,85 @@
...
+   * @return the message payload for this event
+   */
+  Message getMessage();
+
+  /**
+   * Returns the authentication information for the event
+   * 
+   * @return authentication information for the event, may be {@link Optional#empty()}.
+   */
+  Optional<Authentication> getAuthentication();
```
dfeist: Did you see if this could be moved to EventContext? (if it makes sense but is not trivial, we can do it in another PR).

PR #311, opened 2017-08-16
MULE-11065: Make all configuration components AnnotatedObject

elrodro83:


src/main/java/org/mule/runtime/api/lifecycle/Initialisable.java
```diff
@@ -6,13 +6,20 @@
...
+import javax.xml.namespace.QName;
+
 /**
  * {@code Initialisable} is a lifecycle interface that gets called at the initialise lifecycle
  * stage of the implementing service.
  *
  * @since 1.0
  */
-public interface Initialisable {
+public interface Initialisable extends AnnotatedObject {
```
pablolagreca: I don't think we can force all Initialisable objects to be annotated.

pablolagreca: And if we do, then all Startable objects must be AnnotatedObject for the same reason.

pablolagreca: Let's discuss this tomorrow

elrodro83: The difference between `Initialisable` and `Startable`is that the method `initialise()` declares to throw `InitialisationException`, which is constructed with the `Initialisable` instance. I did this before doing the CGLib stuff, maybe we can do that for `Initialisable`s too to avoid declaring this here.

PR #315, opened 2017-08-21
MULE-13263: Interception API no calling before & around if the processor

elrodro83:
fails


src/main/java/org/mule/runtime/api/interception/ProcessorParameterValue.java
```diff
@@ -0,0 +1,39 @@
...
+   * @return the value as a {@link String}, as it is configured in the application configuration, or {@code null} for implicit
+   *         operation parameters.
+   */
+  String providedValue();
+
+  /**
+   * Performs any required expression resolution on {@link #providedValue()}.
+   *
+   * @return if {@link #providedValue()} was an expression, its resolved value. Otherwise, the same as {@link #providedValue()}.
+   * @throws MuleRuntimeException any error that occurs while resolving the value is re-thrown.
```
dfeist: reword.

PR #319, opened 2017-08-24
MULE-13380: Provide a way to unregister ExceptionReader from

elrodro83:
ExceptionHelper


src/main/java/org/mule/runtime/api/exception/ExceptionHelper.java
```diff
@@ -307,4 +307,14 @@ public static void registerExceptionReader(ExceptionReader reader) {
     exceptionReaders.add(reader);
   }
 
+  /**
+   * Registers an exception reader with Mule
```
pablokraan: Javadoc does not match method name/code

PR #331, opened 2017-09-01
MULE-13188: ValueProviderService creates a connection although this m…

estebanwasinger:
…ay not be required

### DIFF
--- a/src/main/java/org/mule/runtime/api/value/ValueProviderService.java
+++ b/src/main/java/org/mule/runtime/api/value/ValueProviderService.java
@@ -7,6 +7,9 @@
 package org.mule.runtime.api.value;
 
 import org.mule.runtime.api.component.location.Location;
+import org.mule.runtime.api.meta.model.parameter.ValueProviderModel;
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/value/ValueProviderService.java
```diff
@@ -23,11 +26,20 @@
...
   ValueResult getValues(Location location, String providerName);
 
+  /**
+   * Retrieves the {@link ValueProviderModel} associated with the given provider name and {@link Location}
+   *
+   * @param location     The {@link Location} where the element is found.
+   * @param providerName The name of the value provider
+   * @return The {@link Optional} {@link ValueProviderModel} associated with the given provider name and {@link Location}
+   */
+  Optional<ValueProviderModel> getModel(Location location, String providerName);
```
marianogonzalez: why would I want the model instead of the valueResult? Isnt' it simpler if the model is part of the ValueResult? Why does this one return Optional while getValues() is not?

PR #332, opened 2017-09-01
MULE-11566 - Rename AnnotatedObject to Component

pablolagreca:


src/main/java/org/mule/runtime/api/meta/AbstractComponent.java
```diff
@@ -8,6 +8,7 @@
 package org.mule.runtime.api.meta;
```
elrodro83: move to component package too?

PR #342, opened 2017-09-08
MULE-13119: Unable to use registered DW functions when extension pref…

alegmarra:
…ix has dash (-)

### DIFF
--- a/src/main/java/org/mule/runtime/api/el/ModuleNamespace.java
+++ b/src/main/java/org/mule/runtime/api/el/ModuleNamespace.java
@@ -6,9 +6,13 @@
  */
 package org.mule.runtime.api.el;
 
-import java.util.Arrays;
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/el/ModuleNamespace.java
```diff
@@ -57,4 +61,22 @@ public boolean equals(Object o) {
...
+    return stream(parts)
+        .filter(StringUtils::isNotBlank)
+        .map(this::normalizePart)
+        .map(StringUtils::capitalize)
+        .collect(toList())
+        .toArray(new String[] {});
+  }
+
+  private String normalizePart(String part) {
+    return stream(part.replaceAll("[^\\w]", " ")
```
marianogonzalez: wouldn't this replace numbers with spaces too?

PR #349, opened 2017-09-13
MULE-13322: Error type not logged when exception is not ME

lucianoRM:


src/main/java/org/mule/runtime/api/exception/MuleException.java
```diff
@@ -100,6 +108,14 @@ private static String repeat(char c, int len) {
     }
   }
 
+  private String getColonMatchingPad(String key) {
+    int padSize = 22 - key.length();
+    if (padSize > 0) {
+      return (repeat(' ', padSize));
```
elrodro83: remove outer parenthesis

src/main/java/org/mule/runtime/api/exception/MuleException.java
```diff
@@ -194,8 +207,10 @@ public String getSummaryMessage() {
     StringBuilder buf = new StringBuilder(1024);
     buf.append(lineSeparator()).append(EXCEPTION_MESSAGE_DELIMITER);
     buf.append("Message               : ").append(message).append(lineSeparator());
-    appendSummaryMessage(buf);
 
+    for (String key : SUMMARY_LOGGING_KEYS) {
+      buf.append(String.format("%s%s: %s\n", key, getColonMatchingPad(key), info.getOrDefault(key, MISSING_DEFAULT_VALUE)));
```
elrodro83: static import

PR #350, opened 2017-09-14
MULE-13545: Cleanup core API

pablokraan:
_ Moving classes out of core's api package
_ When possible, classes where moved to internal
_ Moved to privileged when classes where required on compatibility
_ Moved to other module's API when needed on tests or it made sense
_ Deleted if not needed anymore
_ Some tests were ignored as they use an internal class and need refactoring

### DIFF
--- a/src/main/java/org/mule/runtime/api/connectivity/ConnectivityTestingStrategy.java
+++ b/src/main/java/org/mule/runtime/api/connectivity/ConnectivityTestingStrategy.java
@@ -10,9 +10,9 @@ import org.mule.runtime.api.connection.ConnectionValidationResult;
 
 /**
  * An strategy for doing connectivity testing.
- *
+ * <p>
  * Instances of {@code ConnectivityTestingStrategy} will be discovered through SPI.
- *
+ * <p>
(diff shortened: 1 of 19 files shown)


src/main/java/org/mule/runtime/api/connectivity/ConnectivityTestingStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.connectivity;
+
+import org.mule.runtime.api.connection.ConnectionValidationResult;
+
+/**
+ * An strategy for doing connectivity testing.
+ *
```
elrodro83: `<p>`

src/main/java/org/mule/runtime/api/connectivity/ConnectivityTestingStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+ * Instances of {@code ConnectivityTestingStrategy} will be discovered through SPI.
+ *
```
elrodro83: `<p>`

PR #353, opened 2017-09-18
MULE-13225 - Add required product type property in mule-artifact.json file

pablolagreca:


src/main/java/org/mule/runtime/api/deployment/meta/AbstractMuleArtifactModel.java
```diff
@@ -7,40 +7,47 @@
...
    * @param name name of the artifact
    * @param minMuleVersion minimum Mule Runtime version that requires to work correctly.
+   * @param requiredProduct the target product for the artifact
    * @param classLoaderModelLoaderDescriptor describes how to create the class loader for the artifact.
    * @param bundleDescriptorLoader indicates how to load the bundle descriptor.
    */
   protected AbstractMuleArtifactModel(
                                       String name, String minMuleVersion,
-                                      MuleArtifactLoaderDescriptor classLoaderModelLoaderDescriptor,
+                                      Product requiredProduct, MuleArtifactLoaderDescriptor classLoaderModelLoaderDescriptor,
```
gsfernandes: Shouldn't requiredProduct be checked in case if it is null?

PR #363, opened 2017-09-20
MULE-13521: JMS does not communicate properly it's ExternalLibs

estebanwasinger:


src/main/java/org/mule/runtime/api/meta/model/ExternalLibraryModel.java
```diff
@@ -94,6 +94,38 @@ public ExternalLibraryModelBuilder withType(ExternalLibraryType type) {
     }
 
     /**
+     * Suggests Maven coordinates where the required library can be found. This coordinates should
```
marianogonzalez: maven is just a particular use case of this. ~/Program Files/someWindowsThing is a valid suggested coordinate too

src/main/java/org/mule/runtime/api/meta/model/ExternalLibraryModel.java
```diff
@@ -153,6 +187,32 @@ public String getDescription() {
...
+   * {@code artifactId} and {@code version} of the external library.
+   * <p>
+   * By default, the packaging is {@code jar}, so if is required to use a native library, like a .DLL, you will provide:
+   * {@code org.mule.module:a-native-lib:dll:1.0.0} where {@code dll} is the packaging of the library.
+   * <p>
+   * More information in: <a href="https://maven.apache.org/pom.html#Maven_Coordinates">https://maven.apache.org/pom.html#Maven_Coordinates</a>
+   *
+   * @return The optional maven coordinates.
+   */
+  public Optional<String> getCoordinates() {
```
marianogonzalez: suggestedCordinates... cordinatesExample... as this is, it seems to indicate that these are the actual coordinates

src/main/java/org/mule/runtime/api/meta/model/ExternalLibraryModel.java
estebanwasinger: That example is an OS and path based suggestion, something that won't be usable in most scenarios, as it will only work in windows machines and because it is a path of a DLL you can simply add a `@Path` parameter that will fit better.
I don't know if there is another scenario for this feature which the coordinate can be interpreted as another thing.

marianogonzalez: that's the example of a .net based thing. The case is real.

marianogonzalez: I could also use the suggestion as simply `mysql-driver-5.0.jar`

estebanwasinger: > that's the example of a .net based thing. The case is real.

I know that is real, but is not applicable. We can't ensure that the given path will be always valid, for example in mozart won't work (Imagining that we have Mono support) or any non windows computer.

 > I could also use the suggestion as simply mysql-driver-5.0.jar

This scenario is already covered by the `getRegexMatcher` property.

PR #379, opened 2017-10-02
MULE-13690 Adding a new expression language method to enable better l…

machaval:
…ogging

### DIFF
--- a/src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
+++ b/src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
@@ -53,15 +53,14 @@ public interface ExpressionLanguage {
       throws ExpressionExecutionException;
 
   /**
-   * Evaluates an expression according to a given {@link BindingContext}, the global one and the {@link DataType} of the expected
-   * result.
+   * Evaluates an expression according to a given {@link BindingContext} and the global one.
    *
    * The evaluation of this script will do a best effort to avoid failing when the result value can not be represented in the corresponding format.
    *
    * @param expression         the EL expression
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
```diff
@@ -53,6 +53,20 @@
...
   /**
+   * Evaluates an expression according to a given {@link BindingContext}, the global one and the {@link DataType} of the expected
+   * result.
+   *
+   * The evaluation of this script will do a best effort to avoid failing when the result value can not be represented in the corresponding format.
+   *
+   * @param expression         the EL expression
+   * @param context            the current dynamic binding context to consider
+   * @return the result of the expression plus its type
+   * @throws ExpressionExecutionException or during transformation or during transformation
```
afelisatti: throws seems incorrect

src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
```diff
@@ -53,6 +53,20 @@
       throws ExpressionExecutionException;
 
   /**
+   * Evaluates an expression according to a given {@link BindingContext}, the global one and the {@link DataType} of the expected
```
afelisatti: There's no DataType here.

src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
machaval: Why? An exception can still happen.

aiannucci:
jenkins validate

PR #380, opened 2017-10-03
MULE-13648 - Deploying a bug fix version of an application is not causing correct redeploy

pablolagreca:


src/main/java/org/mule/runtime/api/deployment/meta/MuleDeployableModel.java
```diff
@@ -24,7 +24,8 @@
 public abstract class MuleDeployableModel extends AbstractMuleArtifactModel {
 
   private final List<String> configs;
-  private final boolean redeploymentEnabled;
+  // this field must be true by default because it's the default value used when deserializing this class with no content.
```
pablokraan: This value won't be used when an instance with not redeploymentEnabled field is deserialized

PR #393, opened 2017-10-19
MULE-13824 Add ability to evaluate error message lazily when using pr…

dfeist:
…econditions.

### DIFF
--- a/src/main/java/org/mule/runtime/api/util/Preconditions.java
+++ b/src/main/java/org/mule/runtime/api/util/Preconditions.java
@@ -7,6 +7,8 @@
 
 package org.mule.runtime.api.util;
 
+import java.util.function.Supplier;
+
 /**
  * <p>
  * Utility class to validate Preconditions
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/util/Preconditions.java
```diff
@@ -49,4 +72,19 @@ public static void checkState(boolean condition, String message) {
...
 
+  /**
+   * Ensures that an object reference passed as a parameter to the calling method is not null.
+   *
+   * @param reference an object reference
+   * @param errorMessage the exception message supplier to use if the check fails
+   * @return the non-null reference that was validated
+   * @throws NullPointerException if {@code reference} is null
+   */
+  public static <T> T checkNotNull(T reference, Supplier<String> errorMessage) {
```
elrodro83: https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.util.function.Supplier-

elrodro83: remove previous method too

PR #394, opened 2017-10-19
MULE-13828 Avoid resolving notification simple name whenever a notifi…

dfeist:
…cation is instantiated

### DIFF
--- a/src/main/java/org/mule/runtime/api/notification/AbstractServerNotification.java
+++ b/src/main/java/org/mule/runtime/api/notification/AbstractServerNotification.java
@@ -51,8 +51,6 @@ public abstract class AbstractServerNotification extends EventObject implements 
   public static final int NULL_ACTION = 0;
   public static final Object NULL_MESSAGE = "";
 
-  public final String EVENT_NAME = getClass().getSimpleName();
-
   protected String serverId;
 
   protected long timestamp;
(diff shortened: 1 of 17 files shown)


src/main/java/org/mule/runtime/api/notification/AbstractServerNotification.java
```diff
@@ -51,8 +51,6 @@
   public static final int NULL_ACTION = 0;
   public static final Object NULL_MESSAGE = "";
 
-  public final String EVENT_NAME = getClass().getSimpleName();
```
gsfernandes: I think that the best thing to do here is to use FQN and communicate the change to other teams. @canoasan what do think? It is a minor change.

canoasan: Please validate with them first. Let's anticipate any big impact in other teams

gsfernandes: Ok, I'll talk to @wpoch to check if this is possible.

dfeist: Other approach is to use simple name but rather than do this every time do:
```
  public String getEventName(){
    return "MyNoficiation";
  }
```

Goal from performance perspective is simply not to do `getClass().getSimpleName()` for every notification thats created.

wpoch: Personally, I prefer the latest approach. Will be more performant, and allows the inherited class to override the default implementation, and leads to a better design.

PR #407, opened 2017-10-30
MULE-13909: Have Component provide a rootContainerLocation

elrodro83:


src/main/java/org/mule/runtime/api/component/AbstractComponent.java
```diff
@@ -66,4 +70,15 @@ public String getRootContainerName() {
     return rootContainerName;
   }
 
+  @Override
+  public Location getRootContainerLocation() {
+    if (rootContainerLocation == null) {
+      synchronized (rootContainerLocationInitLock) {
```
dfeist: is this lock really needed? worse that can happen is that location gets built and then discarded.

src/main/java/org/mule/runtime/api/component/Component.java
```diff
@@ -57,8 +58,17 @@
    * The actual root container component name. Some components may belong to a template component in the configuration such as a
    * subflow or a module operation. Since those are template that end up being part of another root container component, like a
    * flow, then the root container name may differ from the one on {@link ComponentLocation#getRootContainerName()}.
-   * 
+   *
    * @return the root container component name.
    */
   String getRootContainerName();
```
dfeist: Why not drop this one? (we discussed dropping this with @pablolagreca  already)

elrodro83: this is used in many places. In order to drop this i would have to replace it with Location#toString, which may be expensive since it does a String join internally.

src/main/java/org/mule/runtime/api/component/AbstractComponent.java
elrodro83: Yep, it's just to avoid that

PR #408, opened 2017-11-01
MULE-13890: Chained try elements generate multiple logs for the same error

lucianoRM:


src/main/java/org/mule/runtime/api/exception/MuleException.java
```diff
@@ -38,13 +38,13 @@
   public static final String INFO_LOCATION_KEY = "Element";
   public static final String INFO_SOURCE_XML_KEY = "Element XML";
   public static final String INFO_ERROR_TYPE_KEY = "Error type";
+  public static final String ALREADY_LOGGED_KEY = "Already logged exception";
```
afelisatti: Just `Logged` in case it gets printed.

afelisatti: Also, pattern for name seems to be `INFO_*`

PR #417, opened 2017-11-28
MULE-13774: Add notifications to HTTP request

afelisatti:


src/main/java/org/mule/runtime/api/meta/model/ExtensionModel.java
```diff
@@ -195,4 +197,14 @@
    * @return A {@link Set} of {@link ErrorModel} registered in the extension.
    */
   Set<ErrorModel> getErrorModels();
+
+  /**
+   * @return A {@link Set} of {@link NotificationModel} registered in the extension.
+   * @since 1.1
+   */
+  default Set<NotificationModel> getNotificationModels() {
```
marianogonzalez: don't add a default. Add an entry in api-ignore instead.

src/main/java/org/mule/runtime/api/meta/model/notification/FiresNotifications.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Indicates that the current model can declare which {@link NotificationModel} it fires.
+ *
+ * @since 1.1
+ */
+public interface FiresNotifications {
```
marianogonzalez: I would rename to HasNotifications because 1) is more consistent 2) the model doesn't fire anything, the implementations do

src/main/java/org/mule/runtime/api/meta/model/notification/FiresNotifications.java
```diff
@@ -0,0 +1,28 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.meta.model.notification;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Indicates that the current model can declare which {@link NotificationModel} it fires.
```
marianogonzalez: copy paste the disclaimer I wrote in all the other models sayins that this interface is not to be implemented by the user

src/main/java/org/mule/runtime/api/meta/model/notification/FiresNotifications.java
```diff
@@ -0,0 +1,28 @@
...
+ * @since 1.1
+ */
+public interface FiresNotifications {
+
+  /**
+   * @return a {@link Set} of {@link NotificationModel} with the notifications that the current component fires.
+   *
+   * @see NotificationModel
+   */
+  default Set<NotificationModel> getNotificationModels() {
```
marianogonzalez: no default. Plus, if you were to have a default, this should be emptySet() which is immutable

src/main/java/org/mule/runtime/api/meta/model/notification/ImmutableNotificationModel.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals;
+import static org.apache.commons.lang3.builder.HashCodeBuilder.reflectionHashCode;
+import org.mule.metadata.api.model.MetadataType;
+
+/**
+ * Default and immutable implementation of {@link NotificationModel}
+ *
+ * @since 1.1
+ */
+public class ImmutableNotificationModel implements NotificationModel {
```
marianogonzalez: why is this here instead of being in ext-api like the others?

marianogonzalez: final until proven otherwise

src/main/java/org/mule/runtime/api/meta/model/notification/ImmutableNotificationModel.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private final String namespace;
+  private final String identifier;
+  private final MetadataType metadataType;
+
+  public ImmutableNotificationModel(String namespace, String identifier, MetadataType metadataType) {
```
marianogonzalez: jdoc

src/main/java/org/mule/runtime/api/meta/model/notification/NotificationModel.java
```diff
@@ -0,0 +1,41 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.meta.model.notification;
+
+import org.mule.metadata.api.model.MetadataType;
+import org.mule.runtime.api.meta.model.ComponentModel;
+
+/**
+ * A model which represents a notification that could be fired by the component represented by the {@link ComponentModel}.
+ *
```
marianogonzalez: disclaimer

src/main/java/org/mule/runtime/api/meta/model/notification/NotificationModel.java
```diff
@@ -0,0 +1,41 @@
...
+   * @return the identifier for the notification
+   */
+  String getIdentifier();
+
+  /**
+   * Returns the {@link MetadataType} of the data associated to the notification.
+   *
+   * @return a not {@code null} {@link MetadataType}
+   */
+  MetadataType getDataType();
```
marianogonzalez: getDataType() -> getType()

src/main/java/org/mule/runtime/api/notification/Notification.java
```diff
@@ -43,7 +43,21 @@ default boolean isSynchronous() {
   /**
    * Indicates the kind of action that causes a {@link Notification} to be fired.
    */
-  public interface Action {
+  interface Action {
+
+    /**
+     * @return unique namespace for the action
+     */
+    default String getNamespace() {
```
marianogonzalez: I understand why you did this, but it makes no practical sense. Any Action implementation which doesn't override these methods will fail, not to mention that when you die and meet Liskov in heaven you won't be able to look her in the eye.

src/main/java/org/mule/runtime/api/meta/model/notification/ImmutableNotificationModel.java
afelisatti: because ErrorModel was my prototype

src/main/java/org/mule/runtime/api/meta/model/notification/NotificationModel.java
afelisatti: Done.

PR #418, opened 2017-11-29
MULE-14082: add CRL feature to Mule 4

alepulver:


src/main/java/org/mule/runtime/api/tls/RevocationCheck.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Configures trust store factory parameters for certificate revocation checking according to
+   * the method implementation.
+   *
+   * @param trustStore the trust store configured for the corresponding TLS context
+   * @param defaultTrustAnchors the default trusted CAs from the JVM
+   * @return the configured trust manager factory parameters
+   */
+  ManagerFactoryParameters configFor(KeyStore trustStore, Set<TrustAnchor> defaultTrustAnchors);
```
afelisatti: I don't think this method should be exposed here. What you could do is expose a builder/factory method. I believe that makes more sense than having that in the TlsContextFactoryBuilder.

src/main/java/org/mule/runtime/api/tls/RevocationCheckFactory.java
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Creates a revocation checking mechanism based on standard CRLDP and OCSP extension points in the certificate.
+   *
+   * @param onlyEndEntities only verify last elements of the chain, false by default
+   * @param preferCrls try CRL instead of OCSP first, false by default
+   * @param noFallback do not use the secondary checking method (the one not selected before), false by default
+   * @param softFail avoid verification failure when the revocation server can not be reached or is busy
+   * @return the revocation checking mechanism
+   */
+  RevocationCheck createStandard(Boolean onlyEndEntities, Boolean preferCrls, Boolean noFallback, Boolean softFail);
```
afelisatti: Must this be `Boolean`? What happens when they are `null`?

PR #419, opened 2017-12-01
MULE-14136: Optimize BindingContext creation

elrodro83:


src/main/java/org/mule/runtime/api/el/BindingContextUtils.java
```diff
@@ -53,43 +54,93 @@ private BindingContextUtils() {
...
   }
+
+  /**
+   * Wraps the given supplier {@code s} so that results for a given input are cached.
+   * 
+   * @param s the supplier to memoize
+   * @return the memoized supplier
+   */
+  private static <T> Supplier<T> memoize(Supplier<T> s) {
+    return new Supplier<T>() {
```
marianogonzalez: this is what a LazyValue already does

PR #420, opened 2017-12-05
MULE-14171: SDK loses the Map DataType generics

estebanwasinger:


src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -160,13 +162,26 @@ static Message of(Object payload) {
...
     CollectionBuilder collectionValue(Object[] value);
 
+    /**
+     * Sets the map value for the {@link Message}'s payload to be built.
+     *
+     * @param value     the map content
+     * @param keyType   the map's key type
+     * @param valueType the map's value type
+     * @return this builder
+     * @throws NullPointerException if the content is null
```
afelisatti: `@since`

src/main/java/org/mule/runtime/api/message/Message.java
```diff
@@ -242,37 +257,30 @@ static Message of(Object payload) {
      */
     CollectionBuilder itemMediaType(MediaType mediaType);
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    CollectionBuilder mediaType(MediaType mediaType);
```
afelisatti: Why are you deleting these?

estebanwasinger: they are interface overridings without sense, these are already defined by the `PayloadBuilder`, and mess it up the implementation. Probably Mule PR will be more clarifying

src/main/java/org/mule/runtime/api/message/Message.java
estebanwasinger: Added `@since` in the interface declaration

PR #421, opened 2017-12-06
MULE-14038: Support static custom types

juandesi:


src/main/java/org/mule/runtime/api/metadata/resolving/OutputStaticTypeResolver.java
```diff
@@ -0,0 +1,45 @@
...
+public abstract class OutputStaticTypeResolver implements OutputTypeResolver, StaticResolver {
+
+  /**
+   * {@inheritDoc}
+   * <p>
+   * The category name is not used for {@link StaticResolver}s since they are not registered in the extension.
+   */
+  @Override
+  public String getCategoryName() {
+    return "OUTPUT_" + STATIC_RESOLVER_NAME;
```
estebanwasinger: just for curiosity, if the category name is not used why is being redefined in all the static type resolvers instead of the parent one?

juandesi: If I made the StaticResolver inherit from NamedResolver and implement the getCategory() method there, one implementing InputTypeResolver and StaticResolver there will be 2 getCategory() implementations and java wont allow that.

alegmarra: It is enforced by the NamedTypeResolver, parent of the OutputResolver

src/main/java/org/mule/runtime/api/metadata/resolving/OutputStaticTypeResolver.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * {@link StaticResolver} implementation for the {@link OutputModel}s.
+ * <p>
+ * This {@link OutputTypeResolver} is discarded once the extension model is built. It's not registered with the dynamic
+ * {@link OutputTypeResolver}s.
+ *
+ * @since 1.1
+ */
+public abstract class OutputStaticTypeResolver implements OutputTypeResolver, StaticResolver {
```
alegmarra: what's the benefit of having two abstract classes instead of a single StaticTypeResolver that implements the three interfaces (Input, output, attributes)?

alegmarra: The AttributesStaticResolver is missing too

src/main/java/org/mule/runtime/api/metadata/resolving/StaticResolver.java
```diff
@@ -0,0 +1,33 @@
...
+ * Handles the construction of a custom {@link MetadataType} in any format to be set in a {@link Typed} component statically
+ * when building the extension without going through the metadata service.
+ * <p>
+ * Provided implementations of this interfaces should be used and not this one directly.
+ *
+ * @since 1.1
+ */
+public interface StaticResolver extends NamedTypeResolver {
+
+  String STATIC_RESOLVER_NAME = "STATIC_RESOLVER";
```
alegmarra: check if declaring a default for categoryName here works

src/main/java/org/mule/runtime/api/metadata/resolving/OutputStaticTypeResolver.java
juandesi: to be consistent with all the other interfaces annotatiosn that we have, separating output from input.

src/main/java/org/mule/runtime/api/metadata/resolving/StaticResolver.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.mule.metadata.api.model.MetadataType;
+import org.mule.runtime.api.connection.ConnectionException;
+import org.mule.runtime.api.meta.Typed;
+import org.mule.runtime.api.metadata.MetadataContext;
+import org.mule.runtime.api.metadata.MetadataResolvingException;
+
+/**
+ * Handles the construction of a custom {@link MetadataType} in any format to be set in a {@link Typed} component statically
+ * when building the extension without going through the metadata service.
```
alegmarra: the user of this interface doesn't know about the metadataService. Talk in terms of design time and compile time.

src/main/java/org/mule/runtime/api/metadata/resolving/StaticResolver.java
```diff
@@ -0,0 +1,33 @@
...
+ * <p>
+ * Provided implementations of this interfaces should be used and not this one directly.
```
alegmarra: links

src/main/java/org/mule/runtime/api/metadata/resolving/StaticResolver.java
juandesi: nope, it wont

src/main/java/org/mule/runtime/api/metadata/resolving/StaticResolver.java
juandesi: I don't agree, parent type should not reference implementations.

src/main/java/org/mule/runtime/api/metadata/resolving/InputStaticTypeResolver.java
```diff
@@ -0,0 +1,43 @@
...
+package org.mule.runtime.api.metadata.resolving;
+
+import org.mule.metadata.api.model.MetadataType;
+import org.mule.runtime.api.meta.model.parameter.ParameterModel;
+import org.mule.runtime.api.metadata.MetadataContext;
+
+/**
+ * {@link StaticResolver} implementation for the {@link ParameterModel}s.
+ * <p>
+ * This {@link InputTypeResolver} is discarded once the extension model is built. It's not registered with the dynamic
```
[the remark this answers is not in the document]
juandesi: Its explained in the parent class

src/main/java/org/mule/runtime/api/metadata/resolving/OutputStaticTypeResolver.java
alegmarra: yeah, discard this one, we need the attributes for output

src/main/java/org/mule/runtime/api/metadata/resolving/InputStaticTypeResolver.java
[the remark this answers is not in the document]
alegmarra: fair enough

src/main/java/org/mule/runtime/api/metadata/resolving/StaticResolver.java
alegmarra: I think a see also would be helpful, but :man_shrugging:

juandesi: I was going to put them, but the javadocs already reference child classes so 🤷‍♂️

src/main/java/org/mule/runtime/api/metadata/resolving/OutputStaticTypeResolver.java
marianogonzalez: what's the point of having this auto generated? If i have a context in which there're multiple resolvers, they will all have the same category name. That defeats the purpose.

I think implementing this method should be mandatory. Same thing applies to InputStaticTypeResolver

alegmarra: yeah, totally right, not cool to default it

PR #422, opened 2017-12-11
MULE-14137: Implement Partial fetching in Multilevel DataSense

alegmarra:


src/main/java/org/mule/runtime/api/metadata/resolving/PartialTypeKeysResolver.java
```diff
@@ -0,0 +1,59 @@
...
+   *         possible {@link MetadataResolvingException} reasons
+   * @throws ConnectionException if an error occurs when using the connection provided by the {@link MetadataContext}
+   */
+  MetadataKey resolveChilds(MetadataContext context, T partial) throws MetadataResolvingException, ConnectionException;
+
+  /**
+   * {@inheritDoc}
+   */
+  default String getResolverName() {
+    return "partial_keys";
```
marianogonzalez: This adds more confusion than value

PR #425, opened 2017-12-13
MULE-14167: Listener threads should have the name of the app

elrodro83:


src/main/java/org/mule/runtime/api/scheduler/Scheduler.java
```diff
@@ -67,8 +67,18 @@
...
    */
   String getName();
+
+  /**
+   * @return the string to append to the name of the thread that executes a task for this Scheduler, or {@code null} if nothing is
+   *         to be appended.
+   * 
+   * @since 1.1
+   */
+  default String getThreadNameSuffix() {
```
dfeist: Prefix or Suffix?  It not set shouldn't it be "" rather than "null"?

elrodro83: This suffix is separated from the rest of the trhead name by a `": "`, So empty and null may behabe differently.

PR #428, opened 2017-12-18
MULE-11659: Add tool to ensure that Mule API is backward compatible o…

pablokraan:
…n each release

_ Adding profile to run the Mule module and Revapi plugins only when a Mule module is declared.
_ Adding Revapi exclusions to temporarily ignore broken API


pablokraan:
jenkins validate

PR #432, opened 2017-12-22
MULE-14173: Add a splash information about services when started

elrodro83:


src/main/java/org/mule/runtime/api/service/Service.java
```diff
@@ -22,8 +26,19 @@
...
 public interface Service extends NamedObject {
+
+  /**
+   * Provides a list of lines to show in the splash screen of the Mule Runtime when this service is started.
+   *
+   * @return the lines to show in the splash screen
+   *
+   * @since 1.1
+   */
+  default List<String> splashMessageLines() {
```
pablokraan: Can this method be renamed to something more general? like getServiceDescription? And it could return just a String that contains "\n" to separate it in lines.

pablokraan: Can this method be moved to a new interface and make Service extend from it? (to reuse the concept on different contexts)

src/main/java/org/mule/runtime/api/service/Service.java
```diff
@@ -22,8 +23,12 @@
...
  * to be resolved from the Mule application context. When such invocation is made, those injected parameters cannot be null.
- * 
+ *
  * @since 1.0
  */
-public interface Service extends NamedObject {
+public interface Service extends NamedObject, SplashMessageProvider {
+
+  default String getSplashMessage() {
+    return null;
```
pablokraan: return an empty string

PR #439, opened 2018-01-04
MULE-14292: Add groupCorrelation to DW binding context and SDK Correlation Info

lucianoRM:


src/main/java/org/mule/runtime/api/message/GroupCorrelation.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static java.util.OptionalInt.empty;
+
+import java.io.Serializable;
+import java.util.OptionalInt;
+
+/**
+ * Immutable container for correlation properties relative to a {@link Message}.
+ * 
+ * @since 4.0
```
marianogonzalez: 4.1

src/main/java/org/mule/runtime/api/message/GroupCorrelation.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static java.util.OptionalInt.empty;
+
+import java.io.Serializable;
+import java.util.OptionalInt;
+
+/**
+ * Immutable container for correlation properties relative to a {@link Message}.
+ * 
+ * @since 4.1
```
marianogonzalez: Sorry , my bad. 1.1

PR #441, opened 2018-01-07
MULE-14384: GroupCorrelation move to a new package broke backward

lucianoRM:


src/main/java/org/mule/runtime/api/el/BindingContextUtils.java
```diff
@@ -86,9 +86,9 @@ public static BindingContext addEventBindings(Event event, BindingContext baseCo
     contextBuilder.addBinding(CORRELATION_ID,
                               new LazyValue<>(() -> new TypedValue<>(event.getContext().getCorrelationId(), STRING)));
 
-    contextBuilder.addBinding(GROUP_CORRELATION_INFO,
-                              new LazyValue<>(() -> new TypedValue<>(event.getGroupCorrelation(),
-                                                                     fromType(GroupCorrelation.class))));
+    contextBuilder.addBinding(GROUP_ITEM_SEQUENCE_INFO,
```
dfeist: Not separate `groupItemSeq` in the end?  In most cases DW users will be interested in sequence and not size (and size will also be often not set), so not best to expose this directly?

lucianoRM: I know we discussed that but i gave it some thought and don't know if that is the best approach.
Consider that both: the whole GroupItemSequenceInfo pojo or just the groupSize can be null.
If we had access to the groupSize directly and it returned null, we would not be able to know wether that is because we are receiving an "infinite" group or because it is not an item from a sequence at all (GroupItemSequenceInfo == null).
Yes, we could ask for the sequenceNumber and if it's not null, we know we are within an infinite sequece, but it does not seem intuitive.

dfeist: Make sense, let's just try to shorten name then and remove redundancy. i.e `groupItemSequenceInfo.sequenceNumber` ->   `itemInfo.sequence`/`groupItemInfo.sequence` or similair?

src/main/java/org/mule/runtime/api/message/ItemSequenceInfo.java
```diff
@@ -0,0 +1,78 @@
...
+    this.position = position;
+    this.sequenceSize = sequenceSize.orElse(-1);
+  }
+
+  /**
+   * Gets the position or ordering number for this event in the the sequence.
+   *
+   * @return the position of the item inside the sequence.
+   */
+  public int getPosition() {
```
dfeist: Does this start at 0 or 1? Make it clear in javadoc.

src/main/java/org/mule/runtime/api/message/ItemSequenceInfo.java
```diff
@@ -0,0 +1,78 @@
...
+    return position;
+  }
+
+  /**
+   * Determines how many events are in the sequence
+   *
+   * @return total events in this sequence or null value if the size is not known
+   */
+  public OptionalInt getSequenceSize() {
```
dfeist: is it safe to assume senquence<Integer.MAX_VALUE? Best to use long?

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -71,6 +64,13 @@
    */
   String getCorrelationId();
 
+  /**
+   * Returns the sequence metadata of this event. See {@link ItemSequenceInfo}.
+   *
+   * @return the correlation metadata of this message.
```
dfeist: Remove reference to correlation.

src/main/java/org/mule/runtime/api/event/Event.java
```diff
@@ -71,6 +64,13 @@
    */
   String getCorrelationId();
 
+  /**
+   * Returns the sequence metadata of this event. See {@link ItemSequenceInfo}.
```
dfeist: Explain that this is empty if the event does not form part of a sequence, and has a value if it does form part of a sequence.

src/main/java/org/mule/runtime/api/message/ItemSequenceInfo.java
```diff
@@ -0,0 +1,78 @@
...
+   * @return the position of the item inside the sequence.
+   */
+  public int getPosition() {
+    return position;
+  }
+
+  /**
+   * Determines how many events are in the sequence
+   *
+   * @return total events in this sequence or null value if the size is not known
```
dfeist: Update: it's empty not null if value is not known.

src/main/java/org/mule/runtime/api/message/ItemSequenceInfo.java
lucianoRM: You think? a sequence bigger than Integer.MAX_VALUE? Don't think that a long is necessary.

dfeist: The use case to think about is what happens if a 2-5Gb+ batch file is streamed into a Mule flow and a `<foreach>` or in the future `<split-aggregate>` is used to process each record/line.  Millions of records don't seem a lot in this scenario at all.  Integer is almost certainly big enough for all scenarios.. but if we wanted to be sure... Your call, was just something I considered

PR #446, opened 2018-01-11
MULE-14376: Annotated every element exposed on the API that have usag…

pablokraan:
…e restrictions

### DIFF
--- a/api-changes.json
+++ b/api-changes.json
@@ -7,28 +7,20 @@
           "old": "interface org.mule.runtime.api.tls.TlsContextFactoryBuilder",
           "new": "interface org.mule.runtime.api.tls.TlsContextFactoryBuilder",
           "annotationType": "org.mule.api.annotation.NoImplement",
-          "annotation": "@org.mule.api.annotation.NoImplement",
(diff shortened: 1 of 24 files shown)


api-changes.json
```diff
@@ -66,6 +58,132 @@
...
+          "code": "java.annotation.added",
+          "old": "interface org.mule.runtime.api.meta.model.source.SourceCallbackModel",
+          "new": "interface org.mule.runtime.api.meta.model.source.SourceCallbackModel",
+          "annotationType": "org.mule.api.annotation.NoImplement",
+          "justification": "This interface is not meant to be implemented by the user"
+        },
+        {
+          "code": "java.annotation.added",
+          "old": "interface org.mule.runtime.api.meta.model.stereotype.StereotypeModel",
+          "new": "interface org.mule.runtime.api.meta.model.stereotype.StereotypeModel",
```
marianogonzalez: not to do on this PR, but the plugin should ignore this by default

pablokraan: It can be done... but I'm not sure that it has to be done. Adding any API annotation changes the API, so users should be aware of the change and this file is the right place to list API changes.

PR #458, opened 2018-01-18
MULE-14433: Replace usages of EqualsBuilder#reflectionEquals

alegmarra:


src/main/java/org/mule/runtime/api/metadata/DefaultMetadataKey.java
```diff
@@ -91,16 +90,33 @@ public String getPartName() {
...
   @Override
   public String toString() {
-    return reflectionToString(this);
+    StringBuilder sb = new StringBuilder();
+    return sb
+      .append("{")
+      .append("id: ").append(id)
+      .append("displayName: ").append(displayName)
+      .append("partName: ").append(partName)
+      .append("childs: ").append(childs)
```
afelisatti: What about properties?

PR #463, opened 2018-01-19
MULE-14510: Allows custom schedulers to use a `runInCaller` rejection

elrodro83:
policy

### DIFF
--- a/src/main/java/org/mule/runtime/api/scheduler/SchedulerConfig.java
+++ b/src/main/java/org/mule/runtime/api/scheduler/SchedulerConfig.java
@@ -39,6 +39,7 @@ public class SchedulerConfig {
   private final String schedulerPrefix;
   private final String schedulerName;
   private final Optional<Boolean> waitAllowed;
+  private final Optional<Boolean> runCpuLightWhenTargetBusy;
   private final Supplier<Long> shutdownTimeoutMillis;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/scheduler/SchedulerConfig.java
```diff
@@ -39,22 +39,26 @@ public static SchedulerConfig config() {
...
     this.schedulerPrefix = null;
     this.schedulerName = null;
     this.waitAllowed = empty();
+    this.runCpuLightWhenTargetBusy = empty();
     this.shutdownTimeoutMillis = () -> null;
   }
 
   private SchedulerConfig(Integer maxConcurrentTasks, String schedulerPrefix, String schedulerName,
-                          Optional<Boolean> waitAllowed, Supplier<Long> shutdownTimeoutMillis) {
+                          Optional<Boolean> waitAllowed, Optional<Boolean> runCpuLightWhenTargetBusy,
```
dfeist: Shouldn't this be more generic? `callerRunsOnRejection` or something?

elrodro83: actually, that would cause more confusion because we want that behavior only when going to cpu-light. We want this to be apllicable to the narrowest cases possible.

PR #466, opened 2018-01-25
MULE-14544: Provide a singleton MultiMap instance

elrodro83:


src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -164,6 +164,22 @@ public void putAll(Map<? extends K, ? extends V> aMap) {
...
+   * @param aMultiMap mappings to be stored in this map
+   * @throws ClassCastException if the class of a key or value in the specified map prevents it from being stored in this map
+   * @throws NullPointerException if the specified map is null, or if this map does not permit null keys or values, and the
+   *         specified map contains null keys or values
+   * @throws IllegalArgumentException if some property of a key or value in the specified map prevents it from being stored in
+   *         this map
+   * @since 1.1.1
+   */
+  public void putAll(MultiMap<? extends K, ? extends V> aMultiMap) {
+    aMultiMap.keySet().forEach(key -> put(key, (Collection<V>) aMultiMap.getAll(key)));
```
afelisatti: `keyList`? Haven't compared the performance but might be better.

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -228,4 +244,27 @@ public String toString() {
     }
   }
 
+  @SuppressWarnings("rawtypes")
+  private static final MultiMap EMPTY_MAP = new MultiMap<>().toImmutableMultiMap();
```
afelisatti: This should be at the top.

PR #478, opened 2018-04-06
MULE-14770: Content parameter is recognized as a route when using the aggregators module

lucianoRM:


src/main/java/org/mule/runtime/api/meta/model/nested/HasNestableElementModels.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * A model that contains nestable elements models
+ *
+ * @since 1.2
+ */
+@NoImplement
+public interface HasNestableElementModels {
```
alegmarra: this is a duplicate of `ComposableModel`, except for the Optional which is replaced by getModels().stream().filter().findFirst. I really want the optional sugar, add the default method in the composableModel

src/main/java/org/mule/runtime/api/meta/model/ExtensionModel.java
```diff
@@ -51,7 +52,7 @@
 @NoImplement
 public interface ExtensionModel
     extends NamedObject, DescribedObject, EnrichableModel, HasOperationModels, HasSourceModels, HasFunctionModels,
-    HasConnectionProviderModels, HasDisplayModel, HasExternalLibraries, HasConstructModels {
+    HasConnectionProviderModels, HasDisplayModel, HasExternalLibraries, HasConstructModels, HasNestableElementModels {
```
alegmarra: This is wrong, and ExtensionModel does not contain nested elements. It contains components, and those components can have the nested elements. Flattening this has no sense in the model, please remove

PR #489, opened 2018-06-21
MULE-14913: Allow ExtensionModel to mark a parameter as 'component ID'

alegmarra:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterDeclarer.java
```diff
@@ -118,6 +120,18 @@ public T asConfigOverride() {
     return (T) this;
   }
 
+  /**
+   * Marks the parameter as being a {@link ParameterModel#isComponentId() component ID}.
+   * This also means this is a {@code required} parameter where expressions are {@link ExpressionSupport::NOT_SUPPORTED}.
+   *
+   * @return {@code this} declarer
+   */
+  public T asComponentId() {
```
estebanwasinger: As the Javadoc of `isComponentId` describes, there are a lot of restrictions of what a `Component ID Parameter` could be. I would improve the declarers, to have a `withComponentIdParameter()` that configures the parameter will all the requisites and also prevents to configure it wrong.

alegmarra: this is covered in the ParameterModelValidator. The idea was to avoid having hidden side effects and be explicit to the user that any of the configurations decided are invalid. The only case where we automatically set the restriction is the ExpressionSupport, because it is not the default value. Otherwise, if the user sets the parameter as optional, content, text, etc. we fail with a clear message.

alegmarra: validator: https://github.com/mulesoft/mule-extensions-api/pull/582/files#diff-d4f41e45d10d69e0ae2f750cf3867387R110

PR #491, opened 2018-07-02
MULE-15249: CaseInsensitiveMultiMap#keySet is not case insensitive

elrodro83:


afelisatti:
@elrodro83 check why validations have failed

PR #507, opened 2018-08-29
MULE-14912: Extension model should expose name and config-ref parameters

juandesi:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterizedDeclaration.java
```diff
@@ -52,6 +53,13 @@ public ParameterGroupDeclaration getParameterGroup(String groupName) {
     return parameterGroups.computeIfAbsent(groupName, ParameterGroupDeclaration::new);
   }
 
+  /**
+   * @return the default {@link ParameterGroupDeclaration}, "General".
+   */
```
estebanwasinger: missing `@since`

PR #508, opened 2018-08-29
MULE-15607: Fix computeIfAbsent for JDK9

aiannucci:


src/main/java/org/mule/runtime/api/util/FunctionalUtils.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Utilities for functional programming
+ *
+ * @since 4.0
```
estebanwasinger: Wrong since version

src/main/java/org/mule/runtime/api/util/FunctionalUtils.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Utilities for functional programming
+ *
+ * @since 4.2
```
marianogonzalez: 1.2

PR #524, opened 2018-10-09
MULE-15786: lazyClientCertificate is not serializable in HttpAttributes

lucianoRM:


src/main/java/org/mule/runtime/api/util/LazyValue.java
```diff
@@ -26,9 +26,18 @@
  */
 public class LazyValue<T> implements Supplier<T> {
 
-  private volatile boolean initialized = false;
-  private T value;
-  private Supplier<T> valueSupplier;
+  transient volatile boolean initialized = false;
```
marianogonzalez: these should be protected, not package

src/main/java/org/mule/runtime/api/util/SerializableLazyValue.java
```diff
@@ -0,0 +1,61 @@
...
+    super(supplier);
+  }
+
+  public SerializableLazyValue(T value) {
+    super(value);
+  }
+
+  @Override
+  public T get() {
+    return super.get();
```
marianogonzalez: redundant

src/main/java/org/mule/runtime/api/util/SerializableLazyValue.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Override
+  public <R> R flatMap(Function<T, R> function) {
+    return super.flatMap(function);
```
marianogonzalez: redundant

src/main/java/org/mule/runtime/api/util/LazyValue.java
marianogonzalez: add comment explaining why these must be transient. Is not obvious considering the class isn't serializable

src/main/java/org/mule/runtime/api/util/SerializableLazyValue.java
lucianoRM: it's not redundant because for SerializableLazyValue T extends Serializable. This way it can be declared that get() returns a Serializable object.

src/main/java/org/mule/runtime/api/util/SerializableLazyValue.java
lucianoRM: same as above.

src/main/java/org/mule/runtime/api/util/LazyValue.java
lucianoRM: why?

marianogonzalez: if they're package, they can only be accessed by classes in the same package. If we want to have an EE implementation, you won't be able

src/main/java/org/mule/runtime/api/util/SerializableLazyValue.java
marianogonzalez: but the super T will be inherited anyway. Have you actually ran into an issue removing this?

src/main/java/org/mule/runtime/api/util/LazyValue.java
marianogonzalez: missing

src/main/java/org/mule/runtime/api/util/SerializableLazyValue.java
lucianoRM: As far as i could understand,  T is not inherited. In fact i could name it X and it would still work (As long as the T in the superClass is not bounded).
And since the superClass T is bounded only by Object(Unbounded) and the subClass T is bounded by Serializable and Serializable is bounded by Object, it works. 

Also, yes, i get errors if not overriding this methods because the methods of the superClass are called which return Object, not Serializable, forcing me to cast them.

PR #526, opened 2018-10-12
MULE-15834: Add an API for expression evaluation session

elrodro83:


src/main/java/org/mule/runtime/api/el/ExpressionLanguageSession.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Evaluates an expression considering a set of given bindings passed on construction time.
+ *
+ * @since 1.2
+ */
+public interface ExpressionLanguageSession extends AutoCloseable {
+
+  /**
+   * Evaluates an expression according to a given {@link BindingContext} and the global one.
```
afelisatti: Check this java docs, they reference a BC as a parameter but in this case that's inaccurate.

PR #532, opened 2018-10-23
MULE-15898 Add new ExpressionLanaguageMetadataService

machaval:


src/main/java/org/mule/runtime/api/metadata/ExpressionLanguageMetadataService.java
```diff
@@ -0,0 +1,238 @@
...
+import org.mule.metadata.message.api.MuleEventMetadataType;
+import org.mule.metadata.message.api.MuleEventMetadataTypeBuilder;
+import org.mule.metadata.message.api.el.TypeBindings;
+import org.mule.runtime.api.service.Service;
+
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
```
gsfernandes: javadoc, include since version

PR #533, opened 2018-10-24
MULE-15919 Add new service for querying el capabilities

machaval:


src/main/java/org/mule/runtime/api/el/DataFormat.java
```diff
@@ -0,0 +1,78 @@
...
+package org.mule.runtime.api.el;
+
+import java.util.List;
+
+/**
+ * Describes a supported Data Format (XML, CSV, etc) by the current expression language
+ *
+ * @since 1.2.0
+ */
+public class DataFormat {
```
elrodro83: make this an interface and place this implementation inside DataWeave

src/main/java/org/mule/runtime/api/el/DataFormatConfigOption.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Describes a DataFormat configuration option. Used to configure the reader and writers
+ *
+ * @since 1.2.0
+ */
+public class DataFormatConfigOption {
```
elrodro83: make this an interface and place this implementation inside DataWeave

src/main/java/org/mule/runtime/api/el/ExpressionLanguageCapabilitiesService.java
```diff
@@ -0,0 +1,27 @@
...
+ * @since 1.2.0
+ */
+public interface ExpressionLanguageCapabilitiesService extends Service {
+
+  /**
+   * Returns the list of supported DataFormat
+   *
+   * @return The list of supported data formats
+   */
+  List<DataFormat> getSupportedDataFormats();
```
elrodro83: Do you care about the order? If not, make this a `Set`

PR #540, opened 2018-11-05
[DO NOT MERGE]MULE-14312: Add the capability of mark components as deprecated

ndinu:


src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatableModel.java
```diff
@@ -0,0 +1,32 @@
...
+package org.mule.runtime.api.meta.model.deprecated;
+
+import java.util.Optional;
+
+/**
+ * Interface that indicates that this part of an extension can be deprecated
+ * 
+ * @since 1.2
+ */
+public interface DeprecatableModel {
```
marianogonzalez: Shouldn't this be `DeprecableModel`?

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+package org.mule.runtime.api.meta.model.deprecated;
+
+
+/**
+ * A model that fully describes the deprecation of a part of the extension model that is a {@link DeprecatableModel}.
+ * 
+ * @since 1.2
+ */
+public interface DeprecatedModel {
```
marianogonzalez: Shouldn't this be DeprecationModel?

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * @return a {@link String} that describes why something was deprecated, what can be used as substitute, or both.
+   */
+  String getMessage();
```
marianogonzalez: this should include two `MuleVersion` fields: `since` which should be mandatory, and `removedIn` which should be `Optional`

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatableModel.java
ndinu: yes

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
ndinu: yes

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/FunctionDeclaration.java
```diff
@@ -6,17 +6,23 @@
  */
 package org.mule.runtime.api.meta.model.declaration.fluent;
 
+import org.mule.runtime.api.meta.model.deprecated.DeprecatedModel;
 import org.mule.runtime.api.meta.model.operation.OperationModel;
 
+import java.util.Optional;
+
+import static java.util.Optional.ofNullable;
```
aiannucci: Wrong import ordering

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/StereotypedDeclaration.java
```diff
@@ -6,21 +6,25 @@
  */
 package org.mule.runtime.api.meta.model.declaration.fluent;
 
+import org.mule.runtime.api.meta.model.deprecated.DeprecatedModel;
 import org.mule.runtime.api.meta.model.stereotype.StereotypeModel;
 
 import java.util.List;
+import java.util.Optional;
+
+import static java.util.Optional.ofNullable;
```
aiannucci: Wrong import ordering

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
ndinu: agree

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
ndinu: This is for deprecating things on extensions, I don't think it makes sense to have muleVersions here. Also, why would it matter when was it deprecated?

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatableModel.java
juandesi: I would just name it, `Deprecable`

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/FunctionDeclaration.java
juandesi: 👮

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
ndinu: What if they don't use semantic versioning?

src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/util/DeclarationWalker.java
```diff
@@ -64,81 +50,82 @@ public final void walk(ExtensionDeclaration extensionDeclaration) {
...
    */
   protected void onConnectionProvider(ConnectedDeclaration owner, ConnectionProviderDeclaration declaration) {}
 
   /**
-   * Invoked when an {@link SourceDeclaration} is found in the
-   * traversed {@code extensionDeclaration}
+   * Invoked when an {@link SourceDeclaration} is found in the traversed {@code extensionDeclaration}
    *
-   * @param owner       The declaration that owns the source
+   * @param owner The declaration that owns the source
```
estebanwasinger: don't change all these JDocs

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatableModel.java
marianogonzalez: No. This is inconsistent with the nomenclature. Deprecable would be the name we'd use if this were a behavior interface. All the meta model classes end with Model. Roll this back

src/main/java/org/mule/runtime/api/meta/model/deprecated/DeprecatedModel.java
marianogonzalez: not using semantic versioning is not an option

PR #558, opened 2019-03-14
MULE-16437: Improve ExtensionClient performance

ndinu:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -58,6 +58,11 @@
   public static final String MULE_LOGGING_INTERVAL_SCHEDULERS_LATENCY_REPORT =
       SYSTEM_PROPERTY_PREFIX + "schedulers.latency.report.interval";
 
+  /**
+   * If set, the extension client will not use any cache to reuse resources between calls
+   */
+  public static final String MULE_EXTENSIONS_CLIENT_CACHE_DISABLED = SYSTEM_PROPERTY_PREFIX + "extensionsClient.disableCache";
```
alegmarra: looking at how this property is used in the other PR, given that the value of the property matters, and not just the presence/absence of the property, I'd change the name to `IS_CACHE_DISABLED` and also update the jdoc to talk about the value of the property and not only `if set`.

ndinu: Do you mean something like `IS_MULE_EXTENSIONS_CLIENT_CACHE_DISABLED` and rename the property to mule.extensionsClient.isCacheDisabled ?

PR #562, opened 2019-03-18
MULE-15656: Add new attribute to @ExternalLib annotation

lucianoRM:


gsfernandes:
I would change also one of the extensions for tests and validate that the serialization (json) considers this new field. I don't remember  if there is a custom gson serializer for this.

lucianoRM:
Will handle this on tooling's side

PR #584, opened 2019-07-23
MULE-17240: changed MultiMap constructor to keep input order

facunlk:


src/test/java/org/mule/runtime/api/util/MultiMapTestCase.java
```diff
@@ -20,12 +20,14 @@
...
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.hamcrest.Description;
 import org.hamcrest.TypeSafeMatcher;
+import static org.junit.Assert.assertEquals;
```
elrodro83: static imports go first

src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -80,7 +80,7 @@ public MultiMap(final Map<K, V> parametersMap) {
         values.add(e.getValue());
       }
       return values;
-    }));
+    },(u,v) -> { throw new IllegalStateException(String.format("Duplicate key %s", u)); }, LinkedHashMap::new));
```
elrodro83: format

PR #593, opened 2019-08-21
MULE-17307: MetadataTypes from the extension model have different instances representing the same type

elrodro83:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ExtensionDeclaration.java
```diff
@@ -240,7 +249,9 @@ public ExtensionDeclaration addFunction(FunctionDeclaration function) {
    * @return {@code this} declaration
    */
   public ExtensionDeclaration addType(ObjectType objectType) {
-    types.add(objectType);
+    if (types.add(objectType)) {
```
marianogonzalez: why do you still have the types set? This is unnecesary redundancy that can lead to trouble. WHy don't you just use the valueSet of the map?

PR #601, opened 2019-10-07
MULE-17535: Performance improvements for MultiMap immutability

elrodro83:
* Optimize cloning for immutability
* Add support for unmodifiable wrapping


src/main/java/org/mule/runtime/api/util/MultiMap.java
```diff
@@ -94,6 +122,10 @@ public MultiMap() {
     if (this instanceof ImmutableMultiMap) {
```
eze210: This check could be avoided taking advantage of polymorphism. 
There are other usages of 'this instanceof ...' in this class too.

aiannucci:
Looks like the work of an artist

PR #612, opened 2019-11-15
MULE-17082: Race condition on test connectivity while deploying

fsgonz:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -76,6 +76,14 @@
   public static final String MULE_LOGGING_INTERVAL_SCHEDULERS_LATENCY_REPORT =
       SYSTEM_PROPERTY_PREFIX + "schedulers.latency.report.interval";
 
+
+  /**
+   * This is the timeout in milliseconds to wait before we detect that the test connectivity is done in case the test connectivity
+   * is asynchronously done.
```
marianogonzalez: add since

PR #615, opened 2019-11-19
MULE-15645: Enable flowStack by default

elrodro83:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -51,6 +51,15 @@
    */
   public static final String MULE_STREAMING_MAX_BUFFER_POOL_SIZE = SYSTEM_PROPERTY_PREFIX + "streaming.bufferPoolMaxSize";
 
+  /**
+   * System property key to set the maximum allowed flowStack entries for an event context, before raising a {@code MULE:CRITICAL}
+   * error.
+   * <p>
+   * By default, the value is set to 50.
```
fsgonz: Not sure if the comment about the default should be here as it is not set here.

elrodro83: true, but if eventually the page with the sysprops documentation is extracted form here, this info will be useful

fsgonz: Ok

PR #616, opened 2019-11-25
MULE-17782. Add support for compile method on ExpressionLanguage

machaval:


src/main/java/org/mule/runtime/api/el/CompiledExpression.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  Optional<MediaType> outputType();
+
+
+  /**
+   * Returns the list of external variable references
+   *
+   * @return The list of external variable references
+   */
+  List<ModuleElementName> externalDependencies();
```
marianogonzalez: What's the reason for exposing this concept?

machaval: Several times runtime team has ask me to see if a given script uses a given function to do some optimizations. For example when the lookup function was used. This will expose this. We can remove this

marianogonzalez: nope. Just wanted to know what this is for. It's awesome to finally have this.

PR #620, opened 2019-12-04
MULE-17820: Use SmallMap to store event variables

marianogonzalez:


src/main/java/org/mule/runtime/api/util/CaseInsensitiveMapWrapper.java
```diff
@@ -49,7 +51,14 @@
    * @param map existing map
    */
   public CaseInsensitiveMapWrapper(Map map) {
-    checkArgument(map.isEmpty(), "Cannot create case insensitive map from a non empty one.");
+    this(map, true);
+  }
+
+  private CaseInsensitiveMapWrapper(Map map, boolean requireEmpty) {
+    if (requireEmpty) {
```
elrodro83: `requireNotEmpty`

marianogonzalez: no... It's requires the map to be empty

PR #624, opened 2019-12-27
MULE-17902: Version is not compliant with Semver 2.0

fededonna:
Fix semver calculation in order to be compliant with 2.0 spec

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
         <oldMuleArtifactVersion>1.2.2</oldMuleArtifactVersion>
         <muleRevapiExtensionVersion>1.3.0-SNAPSHOT</muleRevapiExtensionVersion>
         <muleApiAnnotationsVersion>1.2.0-SNAPSHOT</muleApiAnnotationsVersion>
-
+        <semver4jVersion>3.1.0</semver4jVersion>
         <build.helper.maven.plugin.version>3.0.0</build.helper.maven.plugin.version>
     </properties>
 
@@ -198,6 +198,12 @@
             <groupId>com.sun.activation</groupId>
             <artifactId>javax.activation</artifactId>
             <version>${javax.activation.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.vdurmont</groupId>
+            <artifactId>semver4j</artifactId>
+            <version>${semver4jVersion}</version>
         </dependency>
 
         <!-- test dependencies -->
--- a/src/main/java/org/mule/runtime/api/meta/MuleVersion.java
+++ b/src/main/java/org/mule/runtime/api/meta/MuleVersion.java
@@ -10,8 +10,7 @@ import com.google.common.base.Joiner;
 import com.vdurmont.semver4j.Semver;
 import com.vdurmont.semver4j.SemverException;
 
-import java.util.Optional;
-import java.util.regex.Pattern;
+import static com.vdurmont.semver4j.Semver.SemverType.LOOSE;
 
 /**
  * This class represents Mule Software version scheme:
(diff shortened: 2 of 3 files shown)


src/main/java/org/mule/runtime/api/meta/MuleVersion.java
```diff
@@ -25,54 +29,24 @@
...
-      throw new IllegalArgumentException("Invalid version " + version);
-    }
-
     try {
-      major = Integer.parseInt(m.group(1));
-      minor = Integer.parseInt(m.group(3));
-
-      if (m.group(4) != null && m.group(4).startsWith(".")) {
-        revision = Integer.parseInt(m.group(5));
+      this.semver = new Semver(version, Semver.SemverType.LOOSE);
```
marianogonzalez: static import

src/main/java/org/mule/runtime/api/meta/MuleVersion.java
```diff
@@ -163,19 +137,13 @@ public boolean newerThan(MuleVersion version) {
...
-    StringBuilder v = new StringBuilder(major + "." + minor + ".");
-
-    if (revision >= 0) {
-      v.append(revision);
-    } else {
-      v.append("0");
-    }
-
+    StringBuilder v = new StringBuilder(this.semver.getMajor() + "." + semver.getMinor() + "."
+        + Optional.ofNullable(this.semver.getPatch()).orElse(0));
```
marianogonzalez: `Optional` is no replacement for an `if` or a `ternary`

src/main/java/org/mule/runtime/api/meta/MuleVersion.java
```diff
@@ -201,34 +169,34 @@ public int hashCode() {
...
   }
 
   public void setMinor(int minor) {
-    this.minor = minor;
+    this.semver.withIncMinor(-1 * this.semver.getMinor() + minor);
   }
 
   public int getRevision() {
-    return revision;
+    return Optional.ofNullable(this.semver.getPatch()).orElse(NO_REVISION);
```
marianogonzalez: don't use `Optional` as an if

src/main/java/org/mule/runtime/api/meta/MuleVersion.java
```diff
@@ -201,34 +169,34 @@ public int hashCode() {
...
   }
 
   public int getMinor() {
-    return minor;
+    return this.semver.getMinor();
   }
 
   public void setMinor(int minor) {
-    this.minor = minor;
+    this.semver.withIncMinor(-1 * this.semver.getMinor() + minor);
```
marianogonzalez: why not just setting the value directly?

pom.xml
```diff
@@ -200,6 +200,12 @@
             <version>${javax.activation.version}</version>
         </dependency>
 
+        <dependency>
```
marianogonzalez: if you add this dependency here, then you also need to to send PR to `mule-distributions` and `mule-ee-distributions` to update the dependency whitelist. Otherwise, this breaks the distributitons build

fededonna: I will once this gets approved.

src/main/java/org/mule/runtime/api/meta/MuleVersion.java
fededonna: 👍

src/main/java/org/mule/runtime/api/meta/MuleVersion.java
fededonna: because the library does not have this feature. It can not be set and that is why we increase it like this:
```
actual version  - actual version + new version
```

aiannucci:
According to our own definition of semver (https://www.mulesoft.com/legal/versioning-back-support-policy), an optional Service pack number should be allowed. Are we considering this?

marianogonzalez:
> According to our own definition of semver (https://www.mulesoft.com/legal/versioning-back-support-policy), an optional Service pack number should be allowed. Are we considering this?

Our policy might say that, but reality is that several components of our platform don't support 4 digit versions. We've already tried that.

PR #637, opened 2020-02-17
MULE-18063: Need a way to stop built-in schedule for CH apps

szaffarano:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -126,6 +126,11 @@
   public static final String MULE_LIFECYCLE_FAIL_ON_FIRST_DISPOSE_ERROR =
       SYSTEM_PROPERTY_PREFIX + "lifecycle.failOnFirstDisposeError";
 
+  /**
+   * If set to true, schedulers won't start.
```
elrodro83: schedulers -> scheduler flow sources

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -126,6 +126,11 @@
...
+   */
```
elrodro83: `@since 1.3`

PR #638, opened 2020-02-21
MULE-18063: Revert changes

szaffarano:
It revers changes introduced in PR #637 because this feature already exists.  Also, it deprecates the System Property DISABLE_SCHEDULER_SOURCES_PROPERTY and adds this property in the MuleSystemProperties class.

### DIFF
--- a/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
+++ b/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
@@ -109,8 +109,8 @@ public final class MuleSystemProperties {
   public static final String DATA_WEAVE_SCRIPT_PROCESSING_TYPE = SYSTEM_PROPERTY_PREFIX + "dwScript.processingType";
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -127,11 +127,10 @@
...
 
   /**
-   * If set to true, schedulers flow sources won't start.
-   *
-   * @since 1.3.0
+   * This is a configuration property that can be set at deployment time to disable the scheduler message sources to be started
+   * when deploying an application.
    */
-  public static final String MULE_DISABLE_SCHEDULERS = SYSTEM_PROPERTY_PREFIX + "schedulers.disable";
+  public static final String DISABLE_SCHEDULER_SOURCES_PROPERTY = "mule.config.scheduler.disabled";
```
elrodro83: use `SYSTEM_PROPERTY_PREFIX` instead of `"mule."`

szaffarano: done

PR #648, opened 2020-04-03
MULE-18262: Allow to customize the default scheduler fixed frequency …

marianogonzalez:
…through a System Property

### DIFF
--- a/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
+++ b/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
@@ -148,7 +148,7 @@ public final class MuleSystemProperties {
 
   /**
    * Allows to change the default value for the frequency property of the &lt;fixed-frequency&gt; element. Provided value must
-   * be coercible to a {@link Long}
+   * be coercible to a {@link Long} and be expressed in millis.
    *
    * @since 1.3.0
    */
--- a/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
+++ b/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
@@ -8,6 +8,8 @@ package org.mule.runtime.api.util;
 
 import static java.lang.System.getProperty;
 
+import org.mule.runtime.api.streaming.CursorProvider;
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -145,8 +147,18 @@
   public static final String DISABLE_SCHEDULER_SOURCES_PROPERTY = SYSTEM_PROPERTY_PREFIX + "config.scheduler.disabled";
 
   /**
-   * When enabled, this System Property tracks the stacktrace from where the {@link #close()} method was called. It can be used
-   * for troubleshooting purposes (for example, if someone tries to call {@link #openCursor()} on an already closed cursor.
+   * Allows to change the default value for the frequency property of the &lt;fixed-frequency&gt; element. Provided value must
+   * be coercible to a {@link Long}
```
elrodro83, line 151: do we actually want a value provided this way, or just a flag that if set forces the same default as the extension model?

marianogonzalez: I think that it's more helpful and has more applicable use cases this way. Narrowing it down to just the extension model works but might fall short in future scenarios, specially since we still don't know what we want to do with this exactly.

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -145,8 +147,18 @@
   public static final String DISABLE_SCHEDULER_SOURCES_PROPERTY = SYSTEM_PROPERTY_PREFIX + "config.scheduler.disabled";
 
   /**
-   * When enabled, this System Property tracks the stacktrace from where the {@link #close()} method was called. It can be used
-   * for troubleshooting purposes (for example, if someone tries to call {@link #openCursor()} on an already closed cursor.
+   * Allows to change the default value for the frequency property of the &lt;fixed-frequency&gt; element. Provided value must
```
elrodro83, line 150: add in the javadoc that this goes in millis

PR #656, opened 2020-05-19
MULE-18422: `Scheduler` infrastructure type not visible outside of

elrodro83:
runtime


src/main/java/org/mule/runtime/api/scheduler/SchedulingStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.concurrent.ScheduledFuture;
+
+/**
+ * Strategy used by the {@link org.mule.runtime.api.scheduler.Scheduler} for executing jobs.
+ *
+ * @since 4.4
+ */
+@NoImplement
+public interface SchedulingStrategy {
```
marianogonzalez, line 19: I don't think this is a good solution. If you put this here, we still need to then create a facade in `mule-sdk-api` This feels more like a workaround turned API

PR #658, opened 2020-05-20
MULE-18425: Add types support to `ExtensionWalker`

elrodro83:


src/main/java/org/mule/runtime/api/meta/model/util/ExtensionWalker.java
```diff
@@ -251,6 +254,34 @@ private void walkFunctions(HasFunctionModels model) {
     }
   }
 
+  private void walkTypes(ExtensionModel model) {
```
marianogonzalez, line 257: no idempotent version?

src/main/java/org/mule/runtime/api/meta/model/util/IdempotentExtensionWalker.java
```diff
@@ -153,4 +170,68 @@ protected void onOperation(OperationModel model) {}
...
+      }
+      if (getClass() != obj.getClass()) {
+        return false;
+      }
+      MetadataTypeWrapper other = (MetadataTypeWrapper) obj;
+      if (wrapped == null) {
+        if (other.wrapped != null) {
+          return false;
+        }
+      } else if (!wrapped.getAnnotation(TypeIdAnnotation.class)
```
marianogonzalez, line 229: this annotation might not be there, in which case, you need to fallback to wrapped.equals()

src/main/java/org/mule/runtime/api/meta/model/util/IdempotentExtensionWalker.java
```diff
@@ -153,4 +170,68 @@ protected void onOperation(OperationModel model) {}
...
+        return true;
+      }
+      if (obj == null) {
+        return false;
+      }
+      if (getClass() != obj.getClass()) {
+        return false;
+      }
+      MetadataTypeWrapper other = (MetadataTypeWrapper) obj;
+      if (wrapped == null) {
```
marianogonzalez, line 225: wrapped will never be null

src/main/java/org/mule/runtime/api/meta/model/util/IdempotentExtensionWalker.java
```diff
@@ -153,4 +170,68 @@ protected void onOperation(OperationModel model) {}
...
+
+    private final MetadataType wrapped;
+
+    public MetadataTypeWrapper(MetadataType wrapped) {
+      this.wrapped = wrapped;
+    }
+
+    @Override
+    public int hashCode() {
+      return wrapped.hashCode();
```
marianogonzalez, line 210: inconsisten implementation. If the TypeId annotation exists, it should be the hash code of that id. wrapped.hashCode() should only be the fallback

src/main/java/org/mule/runtime/api/meta/model/util/IdempotentExtensionWalker.java
```diff
@@ -153,4 +170,68 @@ protected void onOperation(OperationModel model) {}
...
+   * Wraps a {@link MetadataType} to do a faster equals relying only on the typeId instead of doing a deep cmparison.
+   * 
+   * @since 1.4
+   */
+  private static class MetadataTypeWrapper {
+
+    private final MetadataType wrapped;
+
+    public MetadataTypeWrapper(MetadataType wrapped) {
+      this.wrapped = wrapped;
```
marianogonzalez, line 205: probably worth extracting the typeId annotation at this point

src/main/java/org/mule/runtime/api/meta/model/util/IdempotentExtensionWalker.java
```diff
@@ -153,4 +170,74 @@ protected void onOperation(OperationModel model) {}
...
+        return true;
+      }
+      if (obj == null) {
+        return false;
+      }
+      if (getClass() != obj.getClass()) {
+        return false;
+      }
+      MetadataTypeWrapper other = (MetadataTypeWrapper) obj;
+      if (wrappedTypeId == null) {
```
marianogonzalez, line 231: I'll approve it anyway, but you could have used `Objects.equals(Object, Object)` method here for simplicity

PR #660, opened 2020-05-26
MULE-18041: Extension error handling improvements

IvanAndresFritzler:
This changes aim to fix error type issues where the error thrown by an extension gets replaced by a different one. Prior to this fix, if an extension makes a call to a second one and wraps the potential exception, that error thrown by the "child" extension will replace the one that the "parent" extension is throwing.
Cause has been reduced to the existence of a nested MessagingException due to the repeated processing of the exception by the core. Logic has been added in order to suppress that nested exception. It's message is stored as a new MuleExceptionInfo property ("Caused By"), that will be logged as part of the error only if it has a value.


src/main/java/org/mule/runtime/internal/exception/SuppressedMuleException.java
```diff
@@ -27,12 +27,51 @@
...
+   * The search will stop if a {@link SuppressedMuleException} is found.
+   * @param exception Exception that will be wrapped, suppressing the exception itself or one of it's causes.
+   * @param causeToSuppress Class of the {@link MuleException} that has to be suppressed.
+   * @param logAsCause If true, an entry will be added to the error log, informing the suppressed exception message.
+   * @return {@link SuppressedMuleException} or provided exception if no cause to suppress is found.
    */
-  public SuppressedMuleException(Throwable causeToSuppress) {
-    super(requireNonNull(causeToSuppress, "Cannot suppress a null cause"));
+  public static Throwable suppressIfPresent(Throwable exception, Class<? extends MuleException> causeToSuppress,
+                                            boolean logAsCause) {
```
elrodro83, line 63: `logAsCause` -> in what case wouldn't we want the cause to be logged?

IvanAndresFritzler: Refactoring in order to unify behaviors

src/main/java/org/mule/runtime/api/exception/MuleExceptionInfo.java
```diff
@@ -76,6 +86,14 @@ public void setErrorType(ErrorType errorType) {
     this.errorType = errorType;
   }
 
+  public String getCausedBy() {
+    return causedBy;
+  }
+
+  public void setCausedBy(String causedBy) {
```
elrodro83, line 93: This should contain the errorType as well so it is logged. Why not store the whole exception instead of the cause?

src/main/java/org/mule/runtime/api/exception/ExceptionHelper.java
```diff
@@ -355,8 +363,13 @@ public static Throwable summarise(Throwable t, int depth) {
   public static List<Throwable> getExceptionsAsList(Throwable t) {
     List<Throwable> exceptions = new ArrayList<>(4);
```
marianogonzalez, line 364: why 4? Why not a LinkedLIst?

src/main/java/org/mule/runtime/api/exception/ExceptionHelper.java
```diff
@@ -355,8 +363,13 @@ public static Throwable summarise(Throwable t, int depth) {
...
     Throwable cause = t;
-    while (cause != null && !(cause instanceof SuppressedMuleException)) {
-      exceptions.add(cause);
+    MuleException suppressedCause = null;
+    while (cause != null && !cause.equals(suppressedCause)) {
```
marianogonzalez, line 367: didn't we established that relying on exceptions equals was a very bad idea? Shouldn't you be testing for SupressedException before hand?

src/main/java/org/mule/runtime/api/exception/MuleExceptionInfo.java
IvanAndresFritzler: Made changes in order to be able to log the error type and the message.

src/main/java/org/mule/runtime/api/exception/ExceptionHelper.java
IvanAndresFritzler: Not sure about the reasons behind that number. Will check.

src/main/java/org/mule/runtime/api/exception/ExceptionHelper.java
IvanAndresFritzler: I believe I can replace it with a same object check, no need for equals here.

IvanAndresFritzler: Same applies for the equals check in getRootMuleException

src/main/java/org/mule/runtime/api/exception/ExceptionHelper.java
IvanAndresFritzler: Discussed this with Rodro and it seems that there are performance related reasons for this being an ArrayList of such size.

src/main/java/org/mule/runtime/api/exception/MuleExceptionInfo.java
```diff
@@ -148,4 +182,44 @@ private static String getColonMatchingPad(String key) {
     return "";
   }
 
+  protected static class SuppressedCauses extends ArrayList<MuleException> {
```
elrodro83, line 185: this just overrides `toString`? no need to extend `ArrayList` for that, just provide a helper method.

src/main/java/org/mule/runtime/api/exception/MuleException.java
```diff
@@ -156,11 +157,19 @@ public void addInfo(String name, Object info) {
...
       this.exceptionInfo.setFlowStack((Serializable) info);
+    } else if (INFO_CAUSED_BY_KEY.equals(name)) {
+      this.exceptionInfo.setSuppressedCauses((MuleExceptionInfo.SuppressedCauses) info);
     } else {
       this.exceptionInfo.putAdditionalEntry(name, info);
     }
   }
 
+  public void addInfo(Map<String, Object> info) {
+    for (Map.Entry<String, Object> entry : info.entrySet()) {
```
elrodro83, line 168: `putAll`

src/main/java/org/mule/runtime/api/exception/MuleExceptionInfo.java
IvanAndresFritzler: To be honest, I don't like this class either. At all. I originally wrote it in order to avoid making changes in MuleException#getVerboseMessage() that is happily calling toString(), but I have to make changes there anyway, so I think that I will go ahead and make the necessary in order to remove this class.

src/main/java/org/mule/runtime/api/exception/MuleException.java
elrodro83: why is this marked as resolved? it is still iteration the given map and putting entries individually instead of using putAll method

src/test/java/org/mule/runtime/api/exception/MuleExceptionInfoTestCase.java
```diff
@@ -0,0 +1,71 @@
...
+import org.mule.runtime.api.message.ErrorType;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+
+public class MuleExceptionInfoTestCase {
```
elrodro83, line 21: extend `AbstractMuleTestCase`

src/test/java/org/mule/runtime/api/exception/MuleExceptionTestCase.java
```diff
@@ -0,0 +1,40 @@
...
+package org.mule.runtime.api.exception;
+
+import io.qameta.allure.Issue;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertThat;
+
+public class MuleExceptionTestCase {
```
elrodro83, line 16: extend `AbstractMuleTestCase`

src/test/java/org/mule/runtime/api/internal/exception/SuppressedMuleExceptionTestCase.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.sameInstance;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.collection.IsMapContaining.hasEntry;
+import static org.hamcrest.core.Is.is;
+
+public class SuppressedMuleExceptionTestCase {
```
elrodro83, line 24: extend `AbstractMuleTestCase`

src/main/java/org/mule/runtime/api/exception/MuleException.java
IvanAndresFritzler: Added a code comment to document the reasons as we discussed :)

PR #666, opened 2020-07-14
MULE-18562: Backward compatibility changes

IvanAndresFritzler:


IvanAndresFritzler:
--validate

IvanAndresFritzler:
> wasn't there some ignored tests to be unignored with this change?
This code was not part of the previous review, but I think that the test you are looking for is https://github.com/mulesoft/mule/pull/9046/files#diff-e06a4ac6f04b938a69ef04445f0b7d81

src/main/java/org/mule/runtime/api/message/Error.java
```diff
@@ -72,4 +72,10 @@ default String getFailingComponent() {
    */
   List<Error> getChildErrors();
 
+  /**
+   * List of errors that identified as underlying causes of the parent error.
+   * @return {@link List<Error>} with all the underlying causes.
+   */
+  List<Error> getSuppressedErrors();
```
elrodro83, line 79: `@since`

src/main/java/org/mule/runtime/api/message/Error.java
```diff
@@ -72,4 +72,10 @@ default String getFailingComponent() {
    */
   List<Error> getChildErrors();
 
+  /**
+   * List of errors that identified as underlying causes of the parent error.
```
elrodro83, line 76: which onw would be the parent error in this case? This needs rephrasing

IvanAndresFritzler:
--validate

src/main/java/org/mule/runtime/api/message/Error.java
```diff
@@ -72,4 +72,15 @@ default String getFailingComponent() {
    */
   List<Error> getChildErrors();
 
+  /**
+   * List of errors that where identified as underlying causes of this error.
+   * For instance, the until-successful scope may throw a retry exhausted error including the error that caused the exhaustion as a suppressed error.
+   *
```
elrodro83, line 78: use `<p>` as a paragraph separator in javadocs

PR #671, opened 2020-07-17
MULE-18576: Move ArtifactDeclaration serializer to mule-artifact-declaration and provide method to serialize any public declaration

lucianoRM:
…eclaratio


gsfernandes:
--validate

PR #672, opened 2020-07-20
MULE-17709: Add `error-mapping` information into operations model

elrodro83:


src/main/java/org/mule/runtime/api/meta/model/operation/ErrorMapping.java
```diff
@@ -0,0 +1,26 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.meta.model.operation;
+
+/**
+ * Determines that an error thrown by an operation should be mapped to another.
+ *
+ * @since 1.4
+ */
+public interface ErrorMapping {
```
marianogonzalez, line 14: is this interface really needed as an API?

src/main/java/org/mule/runtime/api/meta/model/operation/ErrorMapping.java
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * @return the type of the error to be mapped from
+   */
+  String getSource();
```
marianogonzalez, line 19: shouldn't this be ErrorTypes rather than strings?

elrodro83:
--validate

PR #673, opened 2020-07-21
MULE-18426: Interception API: Add a callback to source interceptor on…

JuLuzzi:
…ce all event contexts for an event are complete

### DIFF
--- a/src/main/java/org/mule/runtime/api/interception/SourceInterceptor.java
+++ b/src/main/java/org/mule/runtime/api/interception/SourceInterceptor.java
@@ -63,4 +63,17 @@ public interface SourceInterceptor {
    * @param thrown the exception thrown by the intercepted component, if any.
    */
   default void afterCallback(ComponentLocation location, InterceptionEvent event, Optional<Throwable> thrown) {}
+
+  /**
+   * This method is called when the event was already processed and all the event context of that event are completed.
+   * <p>
+   * If the intercepted source callback fails, the {@link #afterTerminated(ComponentLocation, InterceptionEvent, Optional) after}
+   * methods will still be called, with the thrown exception provided in the {@code thrown} parameter.
+   * <p>
+   *
+   * @param location the location and identification properties of the intercepted component in the mule app configuration.
+   * @param event the result of the component.
+   * @param thrown the exception thrown by the intercepted component, if any.
+   */
+  default void afterTerminated(ComponentLocation location, InterceptionEvent event, Optional<Throwable> thrown) {}
 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/interception/SourceInterceptor.java
```diff
@@ -63,4 +63,17 @@ default void beforeCallback(ComponentLocation location, Map<String, ProcessorPar
...
+   * <p>
+   * If the intercepted source callback fails, the {@link #afterTerminated(ComponentLocation, InterceptionEvent, Optional) after}
+   * methods will still be called, with the thrown exception provided in the {@code thrown} parameter.
+   * <p>
+   *
+   * @param location the location and identification properties of the intercepted component in the mule app configuration.
+   * @param event the result of the component.
+   * @param thrown the exception thrown by the intercepted component, if any.
+   */
+  default void afterTerminated(ComponentLocation location, InterceptionEvent event, Optional<Throwable> thrown) {}
```
elrodro83, line 78: looking at this signature, it is unclear what the event/thrown mean, since there can be many because there were many contexts involved. Better replace both with a `org.mule.runtime.api.event.EventContext`

elrodro83, line 78: `@since 4.4`

src/main/java/org/mule/runtime/api/interception/SourceInterceptor.java
```diff
@@ -63,4 +64,13 @@ default void beforeCallback(ComponentLocation location, Map<String, ProcessorPar
    * @param thrown the exception thrown by the intercepted component, if any.
    */
   default void afterCallback(ComponentLocation location, InterceptionEvent event, Optional<Throwable> thrown) {}
+
+  /**
+   * This method is called when the event was already processed and all the event contexts of that event are completed.
```
elrodro83, line 69: add `regardless of whether the events finished successfully, on error, or a combination of both`.

elrodro83, line 69: `when the event was already processed...` -> `when the root {@link EventContext} is terminated` (the remaining should be added as a clarification of what it means for an eventContext to be terminated).

src/main/java/org/mule/runtime/api/interception/SourceInterceptor.java
```diff
@@ -63,4 +64,13 @@ default void beforeCallback(ComponentLocation location, Map<String, ProcessorPar
...
+   *
+   * @param location the location and identification properties of the intercepted component in the mule app configuration.
+   * @param eventContext of the component result.
+   * @since 4.4
+   */
+  default void afterTerminated(ComponentLocation location, EventContext eventContext) {}
```
elrodro83, line 75: `eventContext ` -> `rootEventContext`

PR #676, opened 2020-07-24
MULE-18535: #[message.message.exceptionPayload.info.Element] returns

elrodro83:
null

* Fix for errors in sdk operations

### DIFF
--- a/src/main/java/org/mule/runtime/api/el/BindingContextUtils.java
+++ b/src/main/java/org/mule/runtime/api/el/BindingContextUtils.java
@@ -16,6 +16,8 @@ import static org.slf4j.LoggerFactory.getLogger;
 
 import org.mule.runtime.api.component.location.ComponentLocation;
 import org.mule.runtime.api.event.Event;
+import org.mule.runtime.api.exception.DefaultMuleException;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/el/BindingContextUtils.java
```diff
@@ -207,19 +210,26 @@ public String getName() {
...
     });
 
     private static final long serialVersionUID = -8097230480930728693L;
 
     private final Message message;
-    private transient final Throwable exceptionPayload;
+    private transient final MuleException exceptionPayload;
 
-    public MessageWrapper(Message message, Throwable exceptionPayload) {
+    public MessageWrapper(Message message, Throwable exceptionPayload, String location) {
```
fsgonz, line 224: Shouldn't we also maintain the previous constructor to not break api? Is this a special case?

elrodro83: this class is private

fsgonz: Ok

PR #695, opened 2020-08-19
EE-7419: Provide a way to enable statistics without monitoring

fsgonz:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -172,6 +172,20 @@
    */
   public static final String SHARE_ERROR_TYPE_REPOSITORY_PROPERTY = SYSTEM_PROPERTY_PREFIX + "share.errorTypeRepository";
 
+  /**
+   * When enabled this System Property, the statistics are enabled even if the monitoring service is not acivated.
```
elrodro83, line 176: explain that this property is only read at app deployment time

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -172,6 +172,20 @@
...
+   * 
+   * @since 4.4, 4.3.1
+   */
+  public static final String MULE_ENABLE_STATISTICS = SYSTEM_PROPERTY_PREFIX + "enable.statistics";
+
+  /**
+   * When enabled this System Property, the payload statistics are disabled independently of the statistics flag.
```
elrodro83, line 183: explain that this property is only read at app deployment time

fsgonz:
--validate

fsgonz:
--validate

fsgonz:
--validate

PR #701, opened 2020-08-25
MULE-18573: CursorProvider key collision at CursorProvider cache

mlischetti:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -164,6 +164,13 @@
    */
   public static final String TRACK_CURSOR_PROVIDER_CLOSE_PROPERTY = SYSTEM_PROPERTY_PREFIX + "track.cursorProvider.close";
 
+  /**
+   * When enabled this System Property, more information about streaming will be logged. It can be used for troubleshooting purposes
```
elrodro83, line 168: `When enabled ` -> `When set to @{code true} `

PR #712, opened 2020-09-17
MULE-18776: Add support for SampleDataResult response wrapper

svacas:


gsfernandes:
All of the service provided by Mule have this contract, a result object. Seems that we didn't think about it when defining the API for the org.mule.runtime.core.api.data.sample.SampleDataService. Should we use this on the service @marianogonzalez?
ValueProvider, Metadata, TestConnectivity all of them define a result as the one proposed here by @svacas.
I would also change this in the service. Do you agree?

src/main/java/org/mule/runtime/api/sampledata/ImmutableSampleDataResult.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public Optional<Message> getSampleData() {
+    return ofNullable(message);
+  }
+
+  @Override
+  public Optional<SampleDataFailure> getFailure() {
+    return ofNullable(failure);
```
lucianoRM, line 43: Avoid creating a new Optional instance every time

src/main/java/org/mule/runtime/api/sampledata/SampleDataResult.java
```diff
@@ -0,0 +1,57 @@
...
+    return new ImmutableSampleDataResult(message);
+  }
+
+  /**
+   * Creates  a new {@link SampleDataResult} with a failure result containing the associated {@link SampleDataFailure failure}
+   *
+   * @param failure The generated failure occurred trying to resolve the sample data
+   * @return a {@link SampleDataResult} with the {@link SampleDataFailure}
+   */
+  static SampleDataResult resultFrom(SampleDataFailure failure) {
```
lucianoRM, line 36: Maybe rename these static methods?

If statically imported: `resultFrom()` does not provide much information of which result we are talking about. `sampleDataResultFrom()` would be better. 

If not statically imported: `SampleDataResult.resultFrom()` seems redundant.` from()` would be better: `SampleDataResult.from()`.

lucianoRM (changes requested):

src/main/java/org/mule/runtime/api/sampledata/ImmutableSampleDataResult.java
svacas: you mean caching the optional? I don't see that idiom used anywhere, do you have an example?

src/main/java/org/mule/runtime/api/sampledata/SampleDataResult.java
svacas: It's consistent with `ValueResult` interface as it is. `MetadataResult` uses `success()` and `failure()` not sure if it's better.

src/main/java/org/mule/runtime/api/sampledata/ImmutableSampleDataResult.java
gsfernandes: As coded is right, what is a bad practice is to have a field of type Optional. Caching it won't make any difference here.

gsfernandes (approved):
Just change the PR name and commit message to a Mule ticket instead.

src/main/java/org/mule/runtime/api/sampledata/ImmutableSampleDataResult.java
lucianoRM: The problem is that Optional does not implement Serializable, but we are not using that API anywhere, so I don't know if that is an issue to us. 
I know that you should only use Optionals as return types, but it seems wrong to create the same instance over and over every time the getter methods are called.

src/main/java/org/mule/runtime/api/sampledata/SampleDataResult.java
lucianoRM: You are right. I would love to change all three. But for the sake of consistency lets leave it like this.

src/main/java/org/mule/runtime/api/sampledata/ImmutableSampleDataResult.java
svacas: if it's null it returns the same empty instance but if not you are right. Not sure if the overhead is significant though.

lucianoRM (approved):

PR #731, opened 2020-10-23
MULE-1892: HTTP-SERVICE bug with jdk8u265-b01 and Runtime 4.1.6 in EB…

JuLuzzi:
…S 12.2 Conn

### DIFF
--- a/src/main/java/org/mule/runtime/api/util/IOUtils.java
+++ b/src/main/java/org/mule/runtime/api/util/IOUtils.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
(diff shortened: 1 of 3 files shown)


src/main/java/org/mule/runtime/api/util/IOUtils.java
```diff
@@ -0,0 +1,31 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.util;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Mule API I/O Utilities.
+ * @since 1.2.0
```
aiannucci, line 14: 1.1.6

aiannucci (dismissed):

src/main/java/org/mule/runtime/api/util/IOUtils.java
aiannucci: or 1.1.7?

eze210 (approved):

PR #736, opened 2020-11-10
MULE-18631: Need to be able to set a limit on the number of events to be pushed per poll where N >= 1

ndinu:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -88,6 +88,13 @@
    */
   public static final String DISABLE_SDK_IGNORE_COMPONENT = SYSTEM_PROPERTY_PREFIX + "disableSdkComponentIgnore";
 
+  /**
+   * When present, adds to polling sources the parameter to configure the item limit per poll feature.
```
Bardammu, line 92: I'd say "to configure the number of items per opoll " or something like that, but I'm not Shakespeare

Bardammu (dismissed):

SebaElizalde (approved):

Bardammu (approved):

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -88,6 +88,13 @@
    */
   public static final String DISABLE_SDK_IGNORE_COMPONENT = SYSTEM_PROPERTY_PREFIX + "disableSdkComponentIgnore";
 
+  /**
+   * When present, adds to polling sources the parameter to configure the maximum number of items per poll to be processed.
```
marianogonzalez, line 92: `When present, adds to polling sources the parameter to configure` -> `When present, Polling Sources are enriched with a parameter for configuring`

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #738, opened 2020-11-11
MULE-18875: ValueProvider/SampleData optional parameters declared in resolver are not included in the model

gabrieldalborgo:


src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -92,6 +97,15 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
     return actingParameters;
```
ndinu, line 97: We should deprecate this method

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -69,15 +71,18 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
    *
    * @since 1.4.0
    */
-  public ValueProviderModel(List<String> actingParameters, boolean requiresConfiguration, boolean requiresConnection,
+  public ValueProviderModel(List<String> actingParameters, List<ParameterModel> parameters, boolean requiresConfiguration,
```
ndinu, line 74: We must create a new constructor, modifying the existing one breaks the api.

ndinu: Maybe the new constructor does not need to receive the actingParameters.

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -46,20 +47,21 @@
...
-   *             {@link ValueProviderModel#ValueProviderModel(java.util.List, boolean, boolean, boolean, java.lang.Integer, java.lang.String, java.lang.String)
+   *             {@link ValueProviderModel#ValueProviderModel(java.util.List, java.util.List, boolean, boolean, boolean, java.lang.Integer, java.lang.String, java.lang.String)
    *             instead}.
    */
   @Deprecated
   public ValueProviderModel(List<String> actingParameters, boolean requiresConfiguration, boolean requiresConnection,
                             boolean isOpen,
                             Integer partOrder, String providerName) {
-    this(actingParameters, requiresConfiguration, requiresConnection, isOpen, partOrder, providerName, providerName);
+    this(actingParameters, emptyList(), requiresConfiguration, requiresConnection, isOpen, partOrder, providerName, providerName);
```
ndinu, line 57: keep delegating on the same constructor

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
gabrieldalborgo: But this constructor had been shipped in a released version?

ndinu: Good point, I did not see that it has the @since 1.4.0 there. I still think it is better to create a new one though we would not be breaking semver by changing it.

gabrieldalborgo: Regarding 'actingParameters', do you think that we must infer them from the list of ParameterModel?
I did not do that because this class has no bussiness logic, and I preferred to keep it at ValueProvidersParameterDeclarationEnricher class.

ndinu:
LGTM , just needs the sample data part to be added.

gabrieldalborgo:
> LGTM , just needs the sample data part to be added.

@ndinu  I just added the sample data part.

gabrieldalborgo:
--validate

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -29,18 +32,24 @@
...
+   * @param actingParameters      the list of parameters that are required to execute the Sample Data Provider resolution
    * @param providerId            the id which unequivocally identifies each provider
    * @param requiresConfiguration indicates if the configuration is required to fetch the sample
    * @param requiresConnection    indicates if the connection is required to fetch the sample
+   *
+   * @deprecated the {@link SampleDataProviderModel} must specify the necessary list of ParameterModel, use
+   *             {@link SampleDataProviderModel#SampleDataProviderModel(String, boolean, boolean, List)}
+   *             instead.
    */
+  @Deprecated
```
marianogonzalez, line 52: this has never been released yet. Just delete it

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -49,18 +58,51 @@ public SampleDataProviderModel(List<String> actingParameters,
...
+   * @param providerId            the id which unequivocally identifies each provider
+   * @param requiresConfiguration indicates if the configuration is required to fetch the sample
+   * @param requiresConnection    indicates if the connection is required to fetch the sample
+   * @param parameters            the list of parameters that the Sample Data Provider takes into account for its resolution
+   */
+  public SampleDataProviderModel(String providerId,
+                                 boolean requiresConfiguration,
+                                 boolean requiresConnection,
+                                 List<ParameterModel> parameters) {
+    requireNonNull(parameters, "'parameters' can't be null");
```
marianogonzalez, line 79: be consistent. Use `checkArgument` too

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -49,18 +58,51 @@ public SampleDataProviderModel(List<String> actingParameters,
...
+    checkArgument(providerId != null && providerId.length() > 0, "providerId cannot be blank");
+
+    this.parameters = parameters;
+    this.actingParameters = parameters.stream().filter(ParameterModel::isRequired).map(ParameterModel::getName).collect(toList());
```
marianogonzalez, line 83: Why do you still need this concept? Optional parameter are also acting, only that optionally. Even if you still need them, this should be done lazily

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -49,18 +58,51 @@ public SampleDataProviderModel(List<String> actingParameters,
...
     this.requiresConnection = requiresConnection;
     this.providerId = providerId;
   }
 
   /**
    * @return the list of parameters that are required to fetch the sample data
+   *
+   * @deprecated Use {@link #getParameters()} instead.
    */
+  @Deprecated
```
marianogonzalez, line 94: delete

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -106,6 +149,7 @@ public boolean equals(Object o) {
   public int hashCode() {
     return new HashCodeBuilder(17, 37)
         .append(actingParameters)
```
marianogonzalez, line 151: remove

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -96,6 +138,7 @@ public boolean equals(Object o) {
 
     return new EqualsBuilder()
         .append(actingParameters, that.actingParameters)
```
marianogonzalez, line 140: remove

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -68,16 +70,55 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
...
+  @Deprecated
+  public ValueProviderModel(List<String> actingParameters, boolean requiresConfiguration,
+                            boolean requiresConnection,
                             boolean isOpen,
                             Integer partOrder, String providerName, String providerId) {
-    checkNotNull(actingParameters, "'actingParameters' can't be null");
-    checkNotNull(partOrder, "'valueParts' can't be null");
-    checkNotNull(providerName, "'providerName' can't be null");
-    checkNotNull(providerId, "'providerId' can't be null");
+    requireNonNull(actingParameters, "'actingParameters' can't be null");
```
marianogonzalez, line 83: why did you made this changes?

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -68,16 +70,55 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
...
+    requireNonNull(partOrder, "'valueParts' can't be null");
+    requireNonNull(providerName, "'providerName' can't be null");
+    requireNonNull(providerId, "'providerId' can't be null");
     this.isOpen = isOpen;
     this.actingParameters = actingParameters;
+    this.parameters = emptyList();
```
marianogonzalez, line 89: This breaches the contract. The class is annotated with `@DoNotInstantiate`. Just delete this

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -68,16 +70,55 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
...
+                            boolean requiresConnection,
+                            boolean isOpen,
+                            Integer partOrder, String providerName, String providerId, List<ParameterModel> parameters) {
+    requireNonNull(parameters, "'parameters' can't be null");
+    requireNonNull(partOrder, "'valueParts' can't be null");
+    requireNonNull(providerName, "'providerName' can't be null");
+    requireNonNull(providerId, "'providerId' can't be null");
+    this.isOpen = isOpen;
+    this.parameters = parameters;
+    this.actingParameters =
```
marianogonzalez, line 120: do this lazily when and if the old getter is ever invoked.

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -149,6 +202,7 @@ public boolean equals(Object o) {
 
     return new EqualsBuilder()
         .append(actingParameters, that.actingParameters)
```
marianogonzalez, line 204: remove

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -162,6 +216,7 @@ public boolean equals(Object o) {
   public int hashCode() {
     return new HashCodeBuilder(17, 37)
         .append(actingParameters)
```
marianogonzalez, line 218: remove

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
gabrieldalborgo: What do you mean by removing this? Delete this constructor? It makes sense because the jdocs says 'since 4.4.0' and it's marked as deprecated.

But, what should I do with the original constructor? To maintain backward compatibility, should I create the ActingParameterModel list from the original acting parameter's List (List <String>)?

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
gabrieldalborgo: Because checkNotNull is deprecated

marianogonzalez: you should use checkArgument instead

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
marianogonzalez: This breaches the contract because you're adding a way of creating an instance in which there's no correlation between the `getParameters()` and `getActingParameter()`methods. Keeping the old constructor would be ideal, but it's simply not possible.

Since the class is annotated with `@NoInstantiate`, you can remove the constructor without actually breaking backwards since we made it part of this class API that the runtime is the only one allowed to instantiated, so you shouldn't have any big issues with that change

marianogonzalez (changes requested):

api-changes.json
```diff
@@ -54,6 +54,16 @@
...
+        },
+        {
+          "code": "java.method.numberOfParametersChanged",
+          "old": "method void org.mule.runtime.api.meta.model.parameter.ValueProviderModel::<init>(java.util.List<java.lang.String>, boolean, boolean, boolean, java.lang.Integer, java.lang.String)",
+          "new": "method void org.mule.runtime.api.meta.model.parameter.ValueProviderModel::<init>(java.util.List<org.mule.runtime.api.meta.model.parameter.ActingParameterModel>, boolean, boolean, boolean, java.lang.Integer, java.lang.String, java.lang.String)",
+          "package": "org.mule.runtime.api.meta.model.parameter",
+          "classSimpleName": "ValueProviderModel",
+          "methodName": "<init>",
+          "elementKind": "constructor",
+          "justification": "The class is annotated with @NoInstantiate"
```
marianogonzalez, line 66: elaborate on this. Explain that keeping this constructor will breach the contract between those two methods

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
```diff
@@ -28,37 +28,37 @@
...
+  public SampleDataProviderModel(List<ActingParameterModel> parameters,
                                  String providerId,
                                  boolean requiresConfiguration,
                                  boolean requiresConnection) {
-    requireNonNull(actingParameters, "'actingParameters' can't be null");
+    checkArgument(parameters != null, "parameters cannot be null");
     checkArgument(providerId != null && providerId.length() > 0, "providerId cannot be blank");
 
-    this.actingParameters = actingParameters;
+    this.parameters = parameters;
```
marianogonzalez, line 51: wrap into an unmodifiableList

src/main/java/org/mule/runtime/api/meta/model/parameter/ActingParameterModel.java
```diff
@@ -0,0 +1,43 @@
...
+import org.mule.api.annotation.NoImplement;
+import org.mule.runtime.api.meta.model.data.sample.SampleDataProviderModel;
+
+/**
+ * An acting parameter of a {@link ValueProviderModel} or {@link SampleDataProviderModel}
+ * <p>
+ * An acting parameter provides a name, a default value and indicates if it is required
+ * </p>
+ *
+ * @since 4.4.0
```
marianogonzalez, line 18: 1.4.0

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -69,15 +49,18 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
...
-    checkNotNull(providerName, "'providerName' can't be null");
-    checkNotNull(providerId, "'providerId' can't be null");
+    checkArgument(parameters != null, "'parameters' can't be null");
+    checkArgument(partOrder != null, "'valueParts' can't be null");
+    checkArgument(providerName != null, "'providerName' can't be null");
+    checkArgument(providerId != null, "'providerId' can't be null");
     this.isOpen = isOpen;
-    this.actingParameters = actingParameters;
+    this.parameters = parameters;
+    this.actingParameters = parameters.stream().filter(ActingParameterModel::isRequired)
```
marianogonzalez, line 62: this should be lazy.

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -87,11 +70,23 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
 
   /**
    * @return the list of parameters that are required to execute the Value Provider resolution.
+   *
+   * @deprecated Use {@link #getParameters()} instead.
```
marianogonzalez, line 74: specify when it was deprecated

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -69,15 +49,18 @@ public ValueProviderModel(List<String> actingParameters, boolean requiresConfigu
...
-    checkNotNull(partOrder, "'valueParts' can't be null");
-    checkNotNull(providerName, "'providerName' can't be null");
-    checkNotNull(providerId, "'providerId' can't be null");
+    checkArgument(parameters != null, "'parameters' can't be null");
+    checkArgument(partOrder != null, "'valueParts' can't be null");
+    checkArgument(providerName != null, "'providerName' can't be null");
+    checkArgument(providerId != null, "'providerId' can't be null");
     this.isOpen = isOpen;
-    this.actingParameters = actingParameters;
+    this.parameters = parameters;
```
marianogonzalez, line 61: wrap into `unmodifiableList`

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
gabrieldalborgo: Doing that causes some tests to fail. Because the serialized extension model will not have the actingParameters, unless you specifically call the getter so that the actingParameters are loaded.

api-changes.json
gabrieldalborgo: Done

src/main/java/org/mule/runtime/api/meta/model/data/sample/SampleDataProviderModel.java
gabrieldalborgo: Done

src/main/java/org/mule/runtime/api/meta/model/parameter/ActingParameterModel.java
gabrieldalborgo: Done

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
gabrieldalborgo: Done

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
gabrieldalborgo: Done

marianogonzalez (approved):

src/main/java/org/mule/runtime/api/meta/model/parameter/ActingParameterModel.java
```diff
@@ -0,0 +1,36 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.meta.model.parameter;
+
+import org.mule.api.annotation.NoImplement;
+import org.mule.runtime.api.meta.model.data.sample.SampleDataProviderModel;
+
+/**
+ * An acting parameter of a {@link ValueProviderModel} or {@link SampleDataProviderModel}
```
marianogonzalez, line 13: An acting parameter of a design time tooling capability, such as VP or SAmple Data... This might eventually be used by other features we're thinking about, or even DataSense itself... We shouldn't have to update the jdoc each time we add a feature

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #748, opened 2020-12-23
MULE-19038: feature flag reserved variable names

szaffarano:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -259,6 +260,15 @@
...
   public static final String MULE_LAX_ERROR_TYPES = SYSTEM_PROPERTY_PREFIX + "errorTypes.lax";
 
+  /**
+   * When set to {@code true} an application won't be able to override reserved properties such as <code>app.name</code>. If it
+   * isn't set, the behaviour will depend on {@link MuleRuntimeFeature#HONOUR_RESERVED_PROPERTIES}.
+   * 
+   * @see MuleRuntimeFeature#HONOUR_RESERVED_PROPERTIES
+   * @since 4.4.0, 4.3.1
+   */
+  public static final String HONOUR_RESERVED_PROPERTIES_PROPERTY = SYSTEM_PROPERTY_PREFIX + "honour.reserved.properties";
```
elrodro83, line 270: overrides should be managed by the feature service internally, in a generic standard way.

elrodro83 (changes requested):

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
szaffarano: I left the System Property constant definition here, but the logic that uses it remains internally to the Feature Flagging mechanism.  Do you think that we should define this constant only [here](https://github.com/mulesoft/mule-api/pull/748/files#diff-9018e4b400b25f6397972e30bf9832135f6557d83869174a7c8a99e87f20f32aR64)?

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -38,7 +41,8 @@
  *       * @since 4.4.0, 4.3.1
  *       *
  *      {@code *}/
- *      HONOUR_RESERVED_PROPERTIES("Whether reserved properties such as app.name can't be overridden by global properties.", "MULE-17659", "4.4.0, 4.3.1"),
+ *      HONOUR_RESERVED_PROPERTIES("Whether reserved properties such as app.name can't be overridden by global properties.",
```
elrodro83, line 44: `app.name` -> `'app.name'`

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -48,20 +52,29 @@
...
-  ;
+
+  /**
+   * An application shouldn't override reserved properties, but it was possible until MULE-17659. This behaviour has to be fixed
+   * by default to any application developed for 4.3.0+ Runtime but can be overridden by setting
+   * {@link MuleSystemProperties#HONOUR_RESERVED_PROPERTIES_PROPERTY} System Property.
+   * 
+   * @since 4.4.0, 4.3.1
+   */
+  HONOUR_RESERVED_PROPERTIES("Whether reserved properties such as app.name can't be overridden by global properties.",
```
elrodro83, line 63: `app.name` -> `'app.name'`

elrodro83 (changes requested):

elrodro83 (dismissed):

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
elrodro83: nah, it's okay to have it here as long as the doc references the proper flag it modifies.

elrodro83 (approved):

PR #781, opened 2021-01-28
MULE-19180: Event internal parameter StackOverflow error while building same event several times

mlischetti:


src/main/java/org/mule/runtime/internal/util/collection/UnmodifiableMap.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+/**
+ * An unmodifiable view of the specified map. 
+ * 
+ * @param <K> the generic type of the keys
+ * @param <V> the generic type of the values
+ * @since 1.4
```
elrodro83, line 23: this should be backported up to 1.2.x

elrodro83 (changes requested):

elrodro83 (approved):

PR #817, opened 2021-04-21
MULE-19041: Support value providers at body level

Bardammu:


src/main/java/org/mule/runtime/api/meta/model/declaration/fluent/ParameterDeclaration.java
```diff
@@ -142,6 +144,14 @@ public void setValueProviderModel(ValueProviderModel valueProviderModel) {
     this.valueProviderModel = valueProviderModel;
   }
 
+  public List<FieldValueProviderModel> getFieldValueProviderModels() {
```
ndinu, line 147: It is fine to add these methods, but to keep using the pattern that is being used for creating the Declarations, the declarer must expose how to add these fieldValueProviderModels

ndinu (changes requested):

src/main/java/org/mule/runtime/api/meta/model/parameter/FieldValueProviderModel.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+
+/**
+ * Model for {@link ParameterModel} to communicate if one of these are capable to provide {@link Value values} for its fields.
+ *
+ * @since 1.4
+ */
+public class FieldValueProviderModel extends ValueProviderModel {
```
marianogonzalez, line 22: Add `@NoInstantiate` and `@NoExtend`

src/main/java/org/mule/runtime/api/meta/model/parameter/FieldValueProviderModel.java
```diff
@@ -0,0 +1,72 @@
...
+
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    FieldValueProviderModel that = (FieldValueProviderModel) o;
+
+    return new EqualsBuilder()
+        .append(targetPath, that.targetPath)
+        .isEquals() && super.equals(o);
```
marianogonzalez, line 64: `super.equals` implementation doens't accept subclasses, therefore this will always return false.  Make sure to always add tests cases when delegating into `super.equals` as an implicit dependency to the super class policy regarding subclasses is created. Need some kind of safeguard to make sure that a change in the base class doesn't generate a behavior change here

src/main/java/org/mule/runtime/api/meta/model/parameter/FieldValueProviderModel.java
```diff
@@ -0,0 +1,72 @@
...
+      return true;
+    }
+
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    FieldValueProviderModel that = (FieldValueProviderModel) o;
+
+    return new EqualsBuilder()
```
marianogonzalez, line 62: use `Objects.equals()` instead

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -26,7 +26,7 @@
  * @since 1.0
  */
 @NoInstantiate
-public final class ValueProviderModel {
```
marianogonzalez, line 29: add `@NoExtend` then

src/main/java/org/mule/runtime/api/value/ValueProviderService.java
```diff
@@ -31,4 +31,17 @@
...
+   *
+   * @param location   The {@link Location} where the element is found.
+   * @param parameter  The name of the parameter to resolve the {@link Value values}.
+   * @param targetPath The path to the field located in the parameter.
+   * @return the {@link ValueResult result} of the resolving of the values.
+   * @see ValueResult
+   *
+   * @since 1.4.0
+   */
+  ValueResult getFieldValues(Location location, String parameter, String targetPath);
```
marianogonzalez, line 46: doesn't this require a `default` impl in order to not break backwards compatibility? surprised that revApi didn't pick this up as an issue

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/value/ValueProviderService.java
ndinu: I think that this has to do with the fact that the interface is annotated with @NoImplement.

ndinu (approved):

marianogonzalez (approved):

PR #819, opened 2021-04-22
MULE-19100 : Provide a reusable mechanism for using object fields as acting parameters

ndinu:


src/main/java/org/mule/runtime/api/meta/model/parameter/ActingParameterModel.java
```diff
@@ -32,4 +32,9 @@
    * @return a boolean value saying if this parameter is required or not
    */
   boolean isRequired();
+
+  /**
+   * @return the path to the acting parameter’s value in the component.
+   */
+  String getPath();
```
marianogonzalez, line 39: not a path. please rename. Also the javadoc should be explicit on the available bindings and illegal uses.

marianogonzalez (changes requested):

PR #822, opened 2021-05-05
MULE-19388: ValueProviderModels from sdk-api usages must provide a SinceMuleVersionModelProperty

ndinu:


src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -166,6 +215,24 @@ public int hashCode() {
...
+    checkArgument(propertyType != null, "Cannot get model properties of a null type");
+    return ofNullable((T) modelProperties.get(propertyType));
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Set<ModelProperty> getModelProperties() {
+    return unmodifiableSet(new LinkedHashSet<>(modelProperties.values()));
```
marianogonzalez, line 236: don't copy on each invocation. Create the copy and the unmodifiable wrapper at the constructor itself

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -67,6 +80,41 @@ public ValueProviderModel(List<ActingParameterModel> parameters, boolean require
...
+    checkArgument(modelProperties != null, "'modelProperties' can't be null");
+    this.isOpen = isOpen;
+    this.parameters = unmodifiableList(parameters);
+    this.requiresConfiguration = requiresConfiguration;
+    this.requiresConnection = requiresConnection;
+    this.partOrder = partOrder;
+    this.providerName = providerName;
+    this.providerId = providerId;
+    this.modelProperties =
+        Arrays.stream(modelProperties).collect(toMap(modelProperty -> modelProperty.getClass(), modelProperty -> modelProperty));
```
marianogonzalez, line 115: static import

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
```diff
@@ -166,6 +217,27 @@ public int hashCode() {
...
+    checkArgument(propertyType != null, "Cannot get model properties of a null type");
+    return ofNullable((T) modelProperties.get(propertyType));
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Set<ModelProperty> getModelProperties() {
+    if (modelPropertiesValues == null) {
```
marianogonzalez, line 238: why the null check? this condition should be impossible

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/meta/model/parameter/ValueProviderModel.java
ndinu: If this is deserialized, wouldn't the modelPropertiesValues be null?

marianogonzalez: that depends on the type adapter impl, but get your point

marianogonzalez (dismissed):

SebaElizalde (approved):

PR #828, opened 2021-06-07
EE-7827: App is unable to resolve mule-scripting

marianorean:


src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -121,7 +122,19 @@
...
-      "4.4.0, 4.3.1", ENABLE_POLICY_ISOLATION_PROPERTY);
+      "4.4.0, 4.3.1", ENABLE_POLICY_ISOLATION_PROPERTY),
+
+  /**
+   * If set to true, the entity resolver will fail proactively if a namespace cannot be resolved, while previously it only failed
+   * when trying to resolve an element from that schema.
+   *
+   * @since 4.4.0
+   */
+  ENTITY_RESOLVER_FAIL_PROACTIVELY(
```
IvanAndresFritzler, line 133: We talked about a behavior that involved retries and I'm not seeing anything about that here. Are you sure that this javadoc reflects the previous behavior and the new one? Consider remove the word "proacrively" I believe it's better to just say "it will fail if a namespace cannot be resolved".

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -121,7 +122,19 @@
...
+          "If set to true, the entity resolver will fail proactively if a namespace cannot be resolved, while previously it only failed when trying to resolve an element from that schema.",
+          "EE-7827",
+          "4.4.0",
+          ENTITY_RESOLVER_FAIL_PROACTIVELY_PROPERTY);
```
IvanAndresFritzler, line 137: Consider replacing "proactively" for a more descriptive phrase (we can think about one together)

IvanAndresFritzler (changes requested):

marianorean:
--validate

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -121,7 +122,19 @@
   ENABLE_POLICY_ISOLATION(
       "If set to true, extensions imported by a policy will be managed in complete isolation from the extensions imported by the application that is being applied to, and validations will prevent the usage of explicit configurations declared by the application as part of the policy initialization.",
       "MULE-19226",
-      "4.4.0, 4.3.1", ENABLE_POLICY_ISOLATION_PROPERTY);
+      "4.4.0, 4.3.1", ENABLE_POLICY_ISOLATION_PROPERTY),
+
+  /**
+   * If set to true, the entity resolver will fail if a namespace cannot be resolved, while previously it only failed after 10
```
IvanAndresFritzler, line 128: Consider this as client facing documentation. Can we be more specific about what is this entity resolver is in charge of. We can think about this together, but I know it's used to parse our XML DSL.

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -323,6 +323,15 @@
   public static final String TO_STRING_TRANSFORMER_TRANSFORM_ITERATOR_ELEMENTS_PROPERTY =
       SYSTEM_PROPERTY_PREFIX + "transformer.toString.transformIteratorElements";
 
+  /**
+   * If set to true, the entity resolver will fail if a namespace cannot be resolved, while previously it only failed after 10
+   * attempts were made.
+   *
+   * @since 4.4.0
```
IvanAndresFritzler, line 330: Are we sure about only Only 4.4.0 here?

IvanAndresFritzler (changes requested):

IvanAndresFritzler (approved):

PR #833, opened 2021-07-14
MULE-19579: NestedComponentModel needs to be parameterizable and composable

marianogonzalez:


src/main/java/org/mule/runtime/api/meta/model/ComponentModelVisitor.java
```diff
@@ -39,4 +42,34 @@
...
+
+  }
+
+  /**
+   * Visits a {@link NestedRouteModel}
+   *
+   * @param model the visited object
+   * @since 1.4.0
+   */
+  default void visit(NestedRouteModel model) {
```
elrodro83, line 72: this iface already has `@NoImplement`, you shouldn't need the default implementations

elrodro83 (changes requested):

elrodro83 (approved):

PR #841, opened 2021-07-22
MULE-19605: Nested elements should also have the concepto of min and …

marianogonzalez:
…max occurs

### DIFF
--- a/src/main/java/org/mule/runtime/api/meta/model/nested/NestableElementModel.java
+++ b/src/main/java/org/mule/runtime/api/meta/model/nested/NestableElementModel.java
@@ -11,6 +11,8 @@ import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.ComponentModelVisitor;
 import org.mule.runtime.api.meta.model.ComposableModel;
 
+import java.util.Optional;
+
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/org/mule/runtime/api/meta/model/nested/NestableElementModel.java
```diff
@@ -23,6 +25,23 @@
...
+   * @return An int greater or equal to zero
+   * @since 1.4.0
+   */
+  int getMinOccurs();
+
+  /**
+   * {@link Optional} value which represents the maximum amount of times that this component can be used inside the owning one.
+   * {@link Optional#empty()} means unbounded.
+   *
+   * @return If present, a number greater or equal than zero.
```
elrodro83, line 40: does a zero make sense for this? should't it validate against that?

elrodro83 (changes requested):

src/main/java/org/mule/runtime/api/meta/model/nested/NestableElementModel.java
marianogonzalez: this is like this to remain consistent with what RouteModel already has

elrodro83 (approved):

PR #889, opened 2021-09-28
The 'application/dw' transformation fails with NullPointerException i…

marianorean:
…n certain scenarios


src/main/java/org/mule/runtime/api/exception/MuleException.java
```diff
@@ -230,7 +230,7 @@ public String getVerboseMessage() {
 
   public String getSummaryMessage() {
     MuleException e = getRootMuleException(this);
-    if (!e.equals(this)) {
+    if (e != null && !e.equals(this)) {
```
elrodro83, line 233: use `Objects.equals for better readability`

elrodro83 (changes requested):
Add a test case for the fixed issue

elrodro83 (approved):

PR #892, opened 2021-10-01
MULE-19815: load components with composite classloader

jcgrethe:


jcgrethe:
--validate

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -150,7 +151,18 @@
...
-      SET_VARIABLE_WITH_NULL_VALUE_PROPERTY);
+      SET_VARIABLE_WITH_NULL_VALUE_PROPERTY),
+
+  /**
+   * If set to true, extensions will be loaded with an artifact classloader.
+   *
+   * @since 4.4.0
+   */
+  START_EXTENSION_COMPONENTS_WITH_ARTIFACT_CLASSLOADER(
+      "When enabled, extensions will be loaded with an artifact classloader.",
```
elrodro83, line 162: replace `an artifact classloader` by a more concrete definition of what artifact's classloader will be used. Remember that artifact may refer to an app, a domain, a plugin, a policy... so this wording is too ambiguous

elrodro83 (changes requested):

elrodro83 (approved):

PR #980, opened 2022-05-27
W-11214361: Provide a basic ComponentDeclaration

elrodro83:


src/main/java/org/mule/runtime/api/component/declaration/ComponentDeclaration.java
```diff
@@ -0,0 +1,97 @@
...
+ * Represents a specific configuration for a concrete {@link ParameterizedModel}.
+ * <p>
+ * Instances are validated during creation, so no instances must contain invalid parameters according to the {@link #getModel()
+ * model}.
+ *
+ * @param <M> the actual {@link ParameterizedModel} sub-type of the declaration.
+ * 
+ * @since 1.5
+ */
+public interface ComponentDeclaration<M extends ParameterizedModel> {
```
marianogonzalez, line 22: We already have an interface with this name. Can you find a different one which better represents the intent of this?

marianogonzalez (changes requested):

PR #983, opened 2022-05-31
W-11117613: Duplicate application of lifecycle phases in Global Error…

marianorean:
… Handlers

### DIFF
--- a/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
+++ b/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
@@ -535,6 +535,14 @@ public final class MuleSystemProperties {
       SYSTEM_PROPERTY_PREFIX + "deployment.validateAppModelWithRegionClassloader";
 
   /**
+   * If set to true, the global error handlers will be reused instead of creating local copies.
+   *
+   * @since 4.5.0, 4.4.1, 4.3.1
+   */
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -534,6 +534,14 @@ public final class MuleSystemProperties {
   public static final String VALIDATE_APPLICATION_MODEL_WITH_REGION_CLASSLOADER_PROPERTY =
       SYSTEM_PROPERTY_PREFIX + "deployment.validateAppModelWithRegionClassloader";
 
+  /**
+   * If set to true, the global error handlers will be reused instead of creating local copies.
+   *
+   * @since 4.5.0, 4.4.1, 4.3.1
+   */
+  public static final String REUSE_GLOBAL_ERROR_HANDLER_PROPERTY =
```
elrodro83, line 542: this is supposed to be a killswitch property, but it named and is being used as a feature flag

elrodro83 (changes requested):

src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
fsgonz: maybe we should use another name but we should probably set it to false as default and only ask the user to enable it in a first attempt. That is my opion. @facunlk

PR #987, opened 2022-06-02
W-10965130: Add system property for new compositeRoutingException log

hyerans:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -550,6 +550,14 @@ public final class MuleSystemProperties {
   public static final String REUSE_GLOBAL_ERROR_HANDLER_PROPERTY =
       SYSTEM_PROPERTY_PREFIX + "reuse.globalErrorHandler";
 
+  /**
+   * If set to true, messages for the exceptions that make up the composite exception won't be provided
+   *
+   * @since 4.5.0
+   */
+  public static final String MULE_PRINT_LEGACY_COMPOSITE_EXCEPTION_LOG =
```
IvanAndresFritzler, line 558: Consider a less ambiguous name. For example, "legacyCompositeRoutingExceptionLog".

Consider a better description. Explain that it affects to the error handling (particularly the logging) of the AbstractForkJoinRouter based Processors, such as the ParallelForEach and the ScatterGather Routers, linking to those classes.

IvanAndresFritzler (changes requested):

PR #996, opened 2022-06-24
W-10877267: A.11.2 Implement the automatic propagation of trace context and baggage

fsgonz:


src/main/java/org/mule/runtime/api/event/DistributedTraceContext.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * @return the map of the baggage items.
+   */
+  Map<String, String> baggageItemsAsMap();
+
+  /**
+   * @return a {@link DistributedTraceContext} that has no fields nor baggage set.
+   */
+  static DistributedTraceContext emptyDistributedEventContext() {
```
marianogonzalez, line 56: why make this API?

fsgonz:
I will make this internal.

PR #999, opened 2022-06-28
W-11302687: Add getAppId and getMinMuleVersion to the ExpressionLangu…

TomoDC:
…ageConfiguration


src/main/java/org/mule/runtime/api/el/DefaultExpressionLanguageConfiguration.java
```diff
@@ -30,4 +39,14 @@ public Charset getDefaultEncoding() {
...
   }
+
+  @Override
+  public String getAppId() {
+    return appId;
+  }
+
+  @Override
+  public Optional<MuleVersion> getMinMuleVersion() {
+    return minMuleVersion;
```
fsgonz, line 50: use ofNullable for not setting the optional directly.

PR #1010, opened 2022-07-06
W-10998630: Remove warning: "An expression value was given for parameter 'variableName' but it doesn't support expressions" from logging

sofiamorseletto:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -545,6 +544,15 @@ public final class MuleSystemProperties {
   public static final String MULE_PRINT_DETAILED_COMPOSITE_EXCEPTION_LOG_PROPERTY =
       SYSTEM_PROPERTY_PREFIX + "detailedCompositeRoutingExceptionLog";
 
+  /**
+   * When set to true, the variableName identifier in SetVariable is set to support expressions in the Mule Extension Model
+   * (W-10998630)
+   *
+   * @since 4.5.0, 4.4.0-202207, 4.3.0-202207
+   */
+  public static final String SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE_PROPERTY =
```
fsgonz, line 553: rename this to REVERT_SUPPORT_EXPRESSIONS_IN_SET_VARIABLE_PROPERTY, as this indicates that it is a kill switch and not a feature.

fsgonz (changes requested):

fsgonz (approved):

PR #1013, opened 2022-07-15
W-11425552: Add pool strategy in the ByteBufferPoolConfiguration

TomoDC:


src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
```diff
@@ -6,6 +6,8 @@
  */
 package org.mule.runtime.api.el;
 
+import org.mule.runtime.api.lifecycle.Disposable;
+import org.mule.runtime.api.meta.MuleVersion;
```
fsgonz, line 10: unused import.

src/main/java/org/mule/runtime/api/el/ExpressionLanguage.java
```diff
@@ -153,6 +155,15 @@ public String nameIdentifier() {
     };
   }
 
+  /**
+   * Dispose resources needed to be disposed at the DisposePhase.
```
fsgonz, line 159: I would only leave the since, as this is the overridden the method of the extended interface.

src/main/java/org/mule/runtime/api/memory/provider/ByteBufferPoolConfiguration.java
```diff
@@ -35,4 +36,9 @@ public interface ByteBufferPoolConfiguration {
    * @return the max buffer size for the pool
    */
   int getMaxBufferSize();
+
+  /**
+   * @return the pooling strategy to use.
+   */
+  ByteBufferPoolStrategy getPoolStrategy();
```
fsgonz, line 43: rename to getByteBufferPoolStrategy

fsgonz (changes requested):

fsgonz (approved):

fsgonz (approved):

PR #1020, opened 2022-08-10
W-11492717: Polling Source notifications Improvements

SebaElizalde:


SebaElizalde:
--validate

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,71 @@
...
+    registerAction("Item rejected because the source is stopping", ITEM_REJECTED_SOURCE_STOPPING);
+    registerAction("Item rejected due to idempotency", ITEM_REJECTED_IDEMPOTENCY);
+    registerAction("Item rejected due to watermark", ITEM_REJECTED_WATERMARK);
+    registerAction("Item rejected because it exceeded the item limit per poll", ITEM_REJECTED_LIMIT);
+  }
+
+  private String pollId;
+  private String itemId;
+  private Optional<Serializable> watermark;
+  private String eventId;
```
ndinu, line 36: Given that the eventId only makes sense for the ITEM_DISPATCHED notification, shouldn't it have its own notification type? Or should we just leave it as null for the other cases?

SebaElizalde: That is a possibility I considered, but I thought in the end the current implementation is better as a new notification for only 1 notification seemed too much. Current way means less classes (also less listeners that the composer team needs to register) and keeps item notifications grouped together, for only the tradeoff of keeping the parameter null for other cases.

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,71 @@
...
+    registerAction("Item dispatched to flow", ITEM_DISPATCHED);
+    registerAction("Item rejected because the source is stopping", ITEM_REJECTED_SOURCE_STOPPING);
+    registerAction("Item rejected due to idempotency", ITEM_REJECTED_IDEMPOTENCY);
+    registerAction("Item rejected due to watermark", ITEM_REJECTED_WATERMARK);
+    registerAction("Item rejected because it exceeded the item limit per poll", ITEM_REJECTED_LIMIT);
+  }
+
+  private String pollId;
+  private String itemId;
+  private Optional<Serializable> watermark;
```
marianogonzalez, line 35: Bad usage of Optional. Only wrap at the getter level

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static org.mule.runtime.api.notification.PollingSourceNotification.POLL_FAILURE;
+
+import org.mule.api.annotation.Experimental;
+
+import java.io.Serializable;
+import java.util.Optional;
+
+@Experimental
+public class PollingSourceItemNotification extends AbstractServerNotification {
```
marianogonzalez, line 17: jdoc

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public Optional<Serializable> getWatermark() {
+    return watermark;
+  }
+
+  public String getEventId() {
+    return eventId;
+  }
+
+  public void setEventId(String eventId) {
```
marianogonzalez, line 63: shouldn't be mutable

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
marianogonzalez: agree with Nico that null handling should be avoided. This should either be splitted or at the very least use an `Optional` for `getEventId()`. The argument of that increasing the number of listeners doesn't apply as the listeners are bound to the action id, not the notification class and you can always have a base class/interface for polling notifications

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
SebaElizalde: alright, moved to an Optional

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
SebaElizalde: this is needed to later on set the event Id on the notification (see the other PR):
`template.getSourceMessage().getNotification().ifPresent(n -> {
        ((PollingSourceItemNotification) n).setEventId(event.getContext().getId());`

Alternatively, I can propagate the information needed through a Map<String, String> and create the notification at the FlowProcessMediator

marianogonzalez: Not sure I follow the Map part, but the snippet in which you're using this is marked as a no-go in the other PR, so removing that code should also reflect here.

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,79 @@
...
+  public String getPollId() {
+    return pollId;
+  }
+
+  public String getItemId() {
+    return itemId;
+  }
+
+  public Optional<Serializable> getWatermark() {
+    return of(watermark);
```
marianogonzalez, line 64: why not `ofNullable`?

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public Optional<String> getEventId() {
+    return of(eventId);
```
marianogonzalez, line 68: why not `ofNullable`?

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
marianogonzalez: still not removed

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,75 @@
...
+    registerAction("Item rejected due to watermark", ITEM_REJECTED_WATERMARK);
+    registerAction("Item rejected because it exceeded the item limit per poll", ITEM_REJECTED_LIMIT);
+  }
+
+  private String pollId;
+  private String itemId;
+  private Serializable watermark;
+  private String eventId;
+
+  public PollingSourceItemNotification(String pollId, String itemId, Serializable watermark, int action,
```
marianogonzalez, line 46: action should be the first argument

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/notification/PollingSourceItemNotification.java
```diff
@@ -0,0 +1,75 @@
...
+
+  static {
+    registerAction("Item dispatched to flow", ITEM_DISPATCHED);
+    registerAction("Item rejected because the source is stopping", ITEM_REJECTED_SOURCE_STOPPING);
+    registerAction("Item rejected due to idempotency", ITEM_REJECTED_IDEMPOTENCY);
+    registerAction("Item rejected due to watermark", ITEM_REJECTED_WATERMARK);
+    registerAction("Item rejected because it exceeded the item limit per poll", ITEM_REJECTED_LIMIT);
+  }
+
+  private String pollId;
```
marianogonzalez, line 41: these should all be final

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #1038, opened 2022-09-27
W-11791616: Implement OCS support in Mule Framework

rbourbonnavarro:


src/main/java/org/mule/runtime/api/component/ConfigurationProperties.java
```diff
@@ -19,6 +21,16 @@
  */
 public interface ConfigurationProperties {
 
+  /**
+   * Creates a no operation implementation of {@link ConfigurationProperties}.
+   *
+   * @return a no operation implementation of {@link ConfigurationProperties}.
+   * @since 1.5
+   */
+  static ConfigurationProperties nullConfigurationProperties() {
```
elrodro83, line 30: empty instead of null

src/main/java/org/mule/runtime/internal/component/NullConfigurationProperties.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mule.runtime.api.component.ConfigurationProperties;
+
+import java.util.Optional;
+
+/**
+ * Null object implementation for {@link ConfigurationProperties}
+ *
+ * @since 4.5
+ */
+public class NullConfigurationProperties implements ConfigurationProperties {
```
elrodro83, line 20: Empty instead of Null

elrodro83 (changes requested):

PR #1040, opened 2022-09-29
W-11633480: Add evaluateTypeExpression in ExpressionLanguageMetadataS…

TomoDC:
…ervice

### DIFF
--- a/src/main/java/org/mule/runtime/api/metadata/ExpressionLanguageMetadataService.java
+++ b/src/main/java/org/mule/runtime/api/metadata/ExpressionLanguageMetadataService.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/runtime/api/metadata/ExpressionLanguageMetadataService.java
```diff
@@ -147,6 +160,16 @@ ModuleDefinition moduleDefinition(String nameIdentifier, Collection<ModuleDefini
...
 
+  /**
+   * Returns a MetadataType from a given type expression.
+   *
+   * @param typeExpression The type expression to evaluate
+   * @param modules        Other modules available
+   * @return The resolved MetadataType
+   * @since 1.5
+   */
+  Optional<MetadataType> evaluateTypeExpression(String typeExpression, Collection<ModuleDefinition> modules);
```
TomoDC, line 171: This currently returns an Optional, so this can be tested from the mule side. In a next iteration we should work on throwing a descriptive error message instead of returning an optional.

eze210 (approved):

PR #1045, opened 2022-10-11
W-11707589: Add MetadataTypes to ExpressionModule

TomoDC:


src/main/java/org/mule/runtime/api/el/ExpressionModule.java
```diff
@@ -55,11 +57,13 @@ static ExpressionModule.Builder builder(ModuleNamespace namespace) {
 
   /**
    * Returns the namespace of this module
-   * 
+   *
    * @return The namespace
    */
   ModuleNamespace namespace();
 
+  List<MetadataType> declaredTypes();
```
eze210, line 65: Add jdoc

eze210 (changes requested):

eze210 (approved):

PR #1054, opened 2022-11-10
W-11300058: ResolverSetUtils must be able to resolve SubtypeMapping values

ndinu:


src/main/java/org/mule/runtime/api/parameterization/value/ArrayValueDeclarer.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Describes an Array value to be created.
+ *
+ * This should be used the same way regardless of its actual format (Java object, Json/XML document, CSV line, etc).
+ * 
+ * @since 1.5.0
+ */
+@NoImplement
+public interface ArrayValueDeclarer {
```
marianogonzalez, line 21: add `@Experimental` on all the classes created in this PR

src/main/java/org/mule/runtime/api/parameterization/AbstractComponentParameterizationBuilderFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import static java.lang.String.format;
+import static java.util.ServiceLoader.load;
+
+import org.mule.api.annotation.NoExtend;
+import org.mule.runtime.api.meta.model.parameter.ParameterizedModel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@NoExtend
+public abstract class AbstractComponentParameterizationBuilderFactory {
```
marianogonzalez, line 18: this should ideally not live in this module, and that the very least, not be API

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/parameterization/AbstractComponentParameterizationBuilderFactory.java
ndinu: This is the interface that is exposed so that mule implements it and exposes the implementation via spi. Shouldn't it be part of the api?

src/main/java/org/mule/runtime/api/parameterization/value/ArrayValueDeclarer.java
ndinu: I have just added this.

src/main/java/org/mule/runtime/api/parameterization/AbstractComponentParameterizationBuilderFactory.java
marianogonzalez: I see... not a fan of this pattern but let's keep that out of the scope of this PR

src/main/java/org/mule/runtime/api/parameterization/AbstractComponentParameterizationBuilderFactory.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.mule.api.annotation.Experimental;
+import org.mule.api.annotation.NoExtend;
+import org.mule.runtime.api.meta.model.parameter.ParameterizedModel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@NoExtend
+@Experimental
+public abstract class AbstractComponentParameterizationBuilderFactory {
```
marianogonzalez, line 20: jdoc

marianogonzalez (changes requested):

PR #1062, opened 2022-11-16
W-12073186: Enable the possibility of parallel extension model loading when deploying applications

marianogonzalez:
…g when deploying applications


marianogonzalez:
--validate

elrodro83 (approved):

PR #1074, opened 2022-12-14
W-12181023: Implement evict metadata cache/storage APIs (#1073)

mbuchwald:


src/main/java/org/mule/runtime/api/metadata/MetadataStorage.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Like {@link MetadataCache}, this component provides the capability to store data so future requests for that data can be served
+ * faster and accessed from every {@link InputTypeResolver}, {@link TypeKeysResolver} and {@link OutputTypeResolver}. It also
+ * allows to evict entries in case recalculations are needed.
+ * 
+ * @since 1.5
+ */
+@NoImplement
+public interface MetadataStorage extends MetadataCache {
```
marianogonzalez, line 23: this shouldn't extend a deprecated interface. If what you want is to maintain some level of compatiblity, the relationship shoudl be inversed

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/metadata/MetadataStorage.java
mbuchwald: This is a cherrypick, Rodro asked me to deprecate the original API

marianogonzalez: let's review the original then. I don't think this should be released like this in 4.5

PR #1103, opened 2023-01-24
W-12422216: Consider application configuration files when creating a Mule Framework from a project

rbourbonnavarro:


src/main/java/org/mule/runtime/api/deployment/meta/MuleDeployableModel.java
```diff
@@ -62,7 +62,8 @@ protected MuleDeployableModel(String name, String minMuleVersion, Product produc
   }
 
   /**
-   * @return the application configuration files
+   * @return the application configuration files, or {@code null} in case the {@link MuleDeployableModel} was created through
```
elrodro83, line 65: this should document what `null` means rather than how it can get into that state

elrodro83 (changes requested):

PR #1116, opened 2023-02-02
W-11151215: Create Logging Information about the resulting MinMuleVersion.

SebaElizalde:


src/main/java/org/mule/runtime/api/meta/MuleVersion.java
```diff
@@ -23,6 +23,7 @@
 public final class MuleVersion {
 
   public static final int NO_REVISION = -1;
+  public static final MuleVersion FIRST_MULE_VERSION = new MuleVersion("4.1.1");
```
marianogonzalez, line 26: remove this constant. As we discussed many times it shouldn't be used (and I didn't saw it used in the code). If you need it as a temporary default, then it shouldn't be in mule-api

marianogonzalez (changes requested):

src/main/java/org/mule/runtime/api/meta/MuleVersion.java
SebaElizalde: moved

PR #1142, opened 2023-05-03
W-10815440: Remove CGLib fallback (kill switch) and dependency from the runtime

mbuchwald:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -526,6 +526,7 @@ public final class MuleSystemProperties {
    * If set to true, the Objects factories will be created with Byte Buddy instead of CGLIB."
    *
    * @since 4.5.0, 4.4.0-202204, 4.3.0-202204
```
elrodro83, line 528: remove 4.5.0

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -309,8 +309,9 @@ public enum MuleRuntimeFeature implements Feature {
    * When enabled, the Objects factories will be created with Byte Buddy instead of CGLIB.
    *
    * @since 4.5.0, 4.4.0-202203, 4.3.0-202203
```
elrodro83, line 311: remove 4.5.0

elrodro83 (changes requested):

PR #1149, opened 2023-05-12
W-13137702: Modularize the embedded module

rbourbonnavarro:


pom.xml
```diff
@@ -154,6 +155,12 @@
             <artifactId>mule-message-metadata-model</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>mule-module-jpms-utils</artifactId>
```
elrodro83, line 160: this module is in mule, this introduces a circular reference between the repos (and an unwanted dependency on all that use mule-api)

elrodro83 (changes requested):

elrodro83 (changes requested):

elrodro83 (approved):

PR #1157, opened 2023-05-22
W-13312344: JPMS - Cannot load services using a utils in a module when

elrodro83:
the interface is in another module


elrodro83:
--validate

elrodro83:
--validate

PR #1165, opened 2023-06-01
W-13465559: Fix revapi validation using as old version 1.4.0 because …

anosenzo:
…we are going to replace 4.5.0-SNAPSHOT with the main branches


elrodro83 (changes requested):
I had already sent this: https://github.com/mulesoft/mule-api/pull/1164

PR #1168, opened 2023-06-02
W-13497626: Add supported java version concept in SDK

marianogonzalez:


elrodro83 (approved):

src/main/java/org/mule/runtime/api/meta/model/ExtensionModel.java
```diff
@@ -219,4 +220,17 @@
...
+   * This is modeled as a set of String in order to accommodate changes in Java versioning, custom vendor schemes or even patch
+   * versions.
+   * <p>
+   * Items should ideally conform to the versions defined in {@link JavaConstants} but this is not mandatory.
+   *
+   * @return The versions of Java supported by this extension
+   * @see {@link JavaConstants}
+   * @since 1.5.0
+   */
+  Set<String> getSupportedJavaVersions();
```
elrodro83, line 235: maybe the jdoc should clarify what empty means? does it means all are supported or that none are supoorted? Can impls return empty?

elrodro83 (changes requested):

src/main/java/org/mule/runtime/api/meta/model/ExtensionModel.java
marianogonzalez: it will never be empty at this point. If the extension doesn't specify it, it'll assume 8 and 11. I'll add a clarification

elrodro83 (approved):

PR #1188, opened 2023-07-13
W-13753869: Validate deployed extensions support running Java version

marianogonzalez:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -693,6 +693,13 @@ public final class MuleSystemProperties {
   public static final String PARALLEL_EXTENSION_MODEL_LOADING_PROPERTY =
       SYSTEM_PROPERTY_PREFIX + "parallel.extension.model.loading";
 
+  /**
+   * Specifies the policy to validate that all registered extensions are compatible with the Java version Mule is running on
```
elrodro83, line 697: document possible values, default value

elrodro83 (changes requested):

PR #1198, opened 2023-08-10
W-13588773: Replace all javax jars bundled in the runtime with the jakarta equivalent

elrodro83:
* Addendum: replace javax.inject also


elrodro83:
--validate

eze210 (approved):

elrodro83:
--validate

PR #1229, opened 2023-10-04
W-14067135: make Element DSL available in ExceptionPayload

hyerans:


src/main/java/org/mule/runtime/api/el/BindingContextUtils.java
```diff
@@ -301,6 +306,18 @@ public MessageExceptionInfoWrapper(Message message, Throwable exceptionPayload,
       }
     }
 
+    public MessageExceptionInfoWrapper(Message message, Throwable exceptionPayload, String location, String dslSource) {
```
fsgonz, line 309: Integrate the two constructors (this new one with the dslSource and the former one with only one parameter) so that one calls the other so that you don't repeat code.

fsgonz (changes requested):

src/main/java/org/mule/runtime/api/message/Error.java
```diff
@@ -37,6 +37,14 @@ default String getFailingComponent() {
     return null;
   }
 
+  /**
+   * @return the DSL source
+   *
```
fsgonz, line 42: extra line

fsgonz (changes requested):

fsgonz (approved):

PR #1230, opened 2023-10-06
W-14209023: Support in mule-artifact.json of a property that specifies

elrodro83:
which java version the app works for

### DIFF
--- a/src/main/java/org/mule/runtime/api/deployment/meta/MuleDeployableModel.java
+++ b/src/main/java/org/mule/runtime/api/deployment/meta/MuleDeployableModel.java
@@ -8,7 +8,9 @@ package org.mule.runtime.api.deployment.meta;
 
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 
+import static java.util.Collections.emptySet;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mule/runtime/api/deployment/meta/MuleDeployableModel.java
```diff
@@ -87,6 +103,24 @@ public List<String> getSecureProperties() {
...
 
+  /**
+   * Indicates the Java versions this deployable artifact is compatible with.
+   * <p>
+   * This is modeled as a set of String in order to accommodate changes in Java versioning, custom vendor schemes or even patch
+   * versions.
+   * <p>
+   * Items should ideally conform to the versions defined in {@link JavaConstants} but this is not mandatory.
+   * <p>
+   * It may be empty, indication that any Java version is supported.
```
asanguinetti, line 114:
```suggestion
   * It may be empty, indicating that any Java version is supported.
```

asanguinetti (changes requested):
Just a typo.

asanguinetti (approved):

PR #1259, opened 2023-12-26
W-14490182: Set maximum cache size for MediaType to prevent attacks/OOM issue

hyerans:


src/main/java/org/mule/runtime/api/metadata/MediaType.java
```diff
@@ -210,18 +203,19 @@ public static MediaType create(String primaryType, String subType, Charset chars
...
+    return cachedMediaType;
   }
 
   private static MediaType cacheMediaType(final MediaType type, String rfcString, boolean definedInApp) {
-    final MediaType oldValue = cache.putIfAbsent(rfcString, type);
-    return oldValue == null ? type : oldValue;
+    final MediaType oldValue = cache.getIfPresent(rfcString);
+    if (oldValue == null) {
+      cache.put(rfcString, type);
+      cache.cleanUp();
```
fsgonz, line 215: is this needed?

src/test/java/org/mule/runtime/api/test/metadata/MediaTypeTestCase.java
```diff
@@ -344,25 +345,12 @@ public void parseAndGenerateMimeTypeCorrectly() {
   }
 
   @Test
-  public void constantsAndParsedSameInstance() {
```
fsgonz, line 347: Why are this tests removed. Probably the behavior should be the same.

fsgonz (approved):

src/main/java/org/mule/runtime/api/metadata/MediaType.java
hyerans: Without cache.cleanUp() the cache size grows over 32(max size). If we keep using caffeine, we can also consider caffeine policy or scheduler.

src/test/java/org/mule/runtime/api/test/metadata/MediaTypeTestCase.java
hyerans: By evicting cache actively, these sameInstance tests can pass equality tests, but we cannot guarantee that we still have the same instance cached.

src/main/java/org/mule/runtime/api/metadata/MediaType.java
ben-manes: You might use `executor(Runnable::run)`, as it otherwise defers eviction to `ForkJoinPool.commonPool()` which adds a bit of a delay that can confuse unit tests. We do that since various callbacks to user code might block for a long time, while the cache's own work is instant. I suspect that's the actual concern, so a direct executor won't make your tests flaky.

A strict maximum size would require serializing writes which limits throughput. Instead Caffeine uses a write buffer to queue up add/update/remove against the eviction policy. That absorbs a burst of writes to replay as a batch without impacting client requests. If the write rate exceeds the threshold (e.g. stress test) then it will block writers instead of allowing runaway growth.

PR #1260, opened 2023-12-26
W-14490182: Set maximum cache size for MediaType

hyerans:
…OM issue (#1259)

* add test

* W-14490182: Set maximum cache size for MediaType to prevent attacks/OOM issue


fsgonz (approved):

fsgonz:
--validate

PR #1275, opened 2024-01-09
W-14608096: timeout is not being considered for local transactions and rollbacks are being silently executed

mbuchwald:


src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
```diff
@@ -473,7 +474,17 @@ public enum MuleRuntimeFeature implements Feature {
...
   DISABLE_SCHEDULER_LOGGING(
       "When enabled, the Scheduler does not log exceptions.",
-      "W-13881167", "4.6.0", DISABLE_SCHEDULER_LOGGING_PROPERTY);
+      "W-13881167", "4.6.0", DISABLE_SCHEDULER_LOGGING_PROPERTY),
+
+  /**
+   * When enabled, when a (local or xa) transaction reached timeout, an error will be thrown that can be handled using error
+   * handling.
+   *
+   * @since 4.6.0
```
elrodro83, line 483: 4.6.1? 4.6.0 is ff already

elrodro83 (dismissed):

src/main/java/org/mule/runtime/api/config/MuleRuntimeFeature.java
elrodro83: still 4.6.0

elrodro83 (changes requested):

elrodro83 (approved):

PR #1308, opened 2024-03-13
W-15247302: Create simple/usual scope and routers OutputResolvers for usability

mbuchwald:


mbuchwald:
--validate

src/main/java/org/mule/runtime/api/metadata/resolving/AllOfRoutesOutputTypeResolver.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * {@link OutputTypeResolver} implementation for Routers.
+ * <p>
+ * This {@link OutputTypeResolver} propagates the metadata the combination of the output of every route of the router. This
+ * symbolizes that the metadata has the result of all the routes.
+ *
+ * @since 1.7
+ */
+public class AllOfRoutesOutputTypeResolver implements OutputTypeResolver<Void> {
```
marianogonzalez, line 23: I don't think these implementations should be API. We need an utils method to get them (probably defined in `mule-core` but open to other idesas) but the impls themselves should ideally be private.

src/main/java/org/mule/runtime/api/metadata/resolving/AllOfRoutesOutputTypeResolver.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override
+  public String getCategoryName() {
+    return "OUTPUT_ROUTER_DYNAMIC";
```
marianogonzalez, line 27: Not a representative name. That been said, should the category and resolverName be parameterizable? Won't we run into issues otherwise?

src/main/java/org/mule/runtime/api/metadata/resolving/OneOfRoutesOutputTypeResolver.java
```diff
@@ -0,0 +1,36 @@
...
+ * This {@link OutputTypeResolver} propagates the metadata the union of the output of every route of the router. This symbolizes
+ * that the metadata can be either of the results of any of the routes.
+ *
+ * @since 1.7
+ */
+public class OneOfRoutesOutputTypeResolver implements OutputTypeResolver<Void> {
+
+  @Override
+  public String getCategoryName() {
+    return "OUTPUT_ROUTER_DYNAMIC";
```
marianogonzalez, line 27: name and category shouldn't mathch with the other resolvers.

src/main/java/org/mule/runtime/api/metadata/resolving/SameAsChainOutputTypeResolver.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mule.runtime.api.metadata.MetadataResolvingException;
+
+/**
+ * {@link OutputTypeResolver} implementation for Scopes.
+ * <p>
+ * This {@link OutputTypeResolver} propagates the metadata result of the inner chain of the scope, without any modification.
+ *
+ * @since 1.7
+ */
+public class SameAsChainOutputTypeResolver implements OutputTypeResolver<Void> {
```
marianogonzalez, line 24: would rename as passthrough

marianogonzalez (changes requested):

PR #1311, opened 2024-03-15
W-15194030: use thread's context class loader for SPI lookup

rbourbonnavarro:
The current approach of searching the services implementations from the class loader of the service contract class worked well because `mule-api` (services contracts) and `mule-core` (services implementations) were always within the same class loader. Now, for the [Multi Mule support](https://docs.google.com/document/d/1Js0hauEpnAEFInOJXrvROkjlALCe5LaHJ2YcHu6DqRE/edit#heading=h.pvww9i63jif) implementation in the Mule Framework that is no longer true, since `mule-core` will be in a class loader which is a child of `mule-api`'s class loader, making this approach non-viable.


rbourbonnavarro:
--validate

PR #1356, opened 2024-06-11
add entryListeners in ObjectStore

hyerans:


pom.xml
```diff
@@ -282,6 +282,12 @@
             <artifactId>json</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>api-annotations</artifactId>
```
eze210, line 287: Why was this dependency added?

src/main/java/org/mule/runtime/api/store/SimpleMemoryObjectStore.java
```diff
@@ -53,6 +57,16 @@ protected T doRemove(String key) {
     return map.remove(key);
   }
 
+  @Override
+  protected UUID doAddEntryListener(ObjectStoreEntryListener listener) {
```
eze210, line 61: You can just implement addEntryListener and removeEntryListener here, and remove the doSomething from everywhere

eze210 (changes requested):

PR #1369, opened 2024-07-01
W-15676027: Increasing visibility of BaseInputMetadataDescriptor and adding InputChainMetadata to ComponentMetadataTypesDescriptor.

asanguinetti:


mbuchwald (dismissed):

mbuchwald (approved):

src/main/java/module-info.java
```diff
@@ -133,6 +133,9 @@
   // for DataWeave
   exports org.mule.runtime.privileged.event;
 
+  // For runtime-extension-model and extensions-support
+  exports org.mule.runtime.privileged.metadata;
```
marianogonzalez, line 137: shouldn't priviledged exports by directed?

src/main/java/org/mule/runtime/privileged/metadata/InternalMetadataContext.java
```diff
@@ -0,0 +1,32 @@
...
+import org.mule.runtime.api.metadata.MetadataContext;
+
+import java.util.Optional;
+
+/**
+ * A {@link MetadataContext} with additional capabilities, for internal use only.
+ *
+ * @since 4.8.0
+ */
+public interface InternalMetadataContext extends MetadataContext {
```
marianogonzalez, line 20: add `@NoImplement` and `@Experimienetal` (with the corresponding jdoc disclaimer)

src/main/java/org/mule/runtime/api/metadata/descriptor/ComponentMetadataTypesDescriptor.java
```diff
@@ -75,6 +75,10 @@ public static class ComponentMetadataTypesDescriptorBuilder {
     private ComponentMetadataTypesDescriptorBuilder() {}
 
     public ComponentMetadataTypesDescriptor.ComponentMetadataTypesDescriptorBuilder withInputMetadataDescriptor(InputMetadataDescriptor inputMetadataDescriptor) {
+      return this.withInputMetadataDescriptor((BaseInputMetadataDescriptor) inputMetadataDescriptor);
```
marianogonzalez, line 78: why cast?

src/main/java/org/mule/runtime/privileged/metadata/InternalMetadataContext.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * @return the {@link ExpressionLanguageMetadataService}, if available.
+   */
+  Optional<ExpressionLanguageMetadataService> getExpressionLanguageMetadataService();
```
marianogonzalez, line 25: why obtain this through the context instead of thorugh `@Inject`? or are we missing DI in the resolvers?

src/main/java/org/mule/runtime/privileged/metadata/InternalMetadataContext.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * @return the {@link TypeBindings} in context, if available.
```
marianogonzalez, line 28: not a helpful jdoc. What is this for? what does it contain?

marianogonzalez (changes requested):

src/main/java/module-info.java
asanguinetti: I thought so too. But it seems we can't do that anymore because of polymules. Because now the target modules will be in a lower layer, and the API layer will not see them.

src/main/java/org/mule/runtime/api/metadata/descriptor/ComponentMetadataTypesDescriptor.java
asanguinetti: because otherwise it will call the same method and it will be an infinite recursion.

src/main/java/org/mule/runtime/privileged/metadata/InternalMetadataContext.java
asanguinetti: I thought it would be enough by just going to the JDoc of `TypeBindings` to understand that. But I can try to add some more data. As to "what is this for" I'll see what I can put without being over-specific.

asanguinetti: Improved, let me know what you think.

src/main/java/org/mule/runtime/privileged/metadata/InternalMetadataContext.java
asanguinetti: We are not injecting the resolvers (checked empirically). Maybe it is possible to add that, but I wouldn’t like to add it to this task's scope.

marianogonzalez (dismissed):

marianogonzalez (dismissed):

marianogonzalez (approved):

PR #1453, opened 2024-10-15
W-17281517: Upgrade jakarta dependencies

marianorean:


src/test/resources/media_type/MediaTypeV10.class
elrodro83, line 1: why is this regenerated? what was thistesting?

elrodro83 (changes requested):

src/test/resources/media_type/MediaTypeV10.class
marianorean: It seems to be testing some kind of serialization compatibility of the class MediaType between V11 (has some attribute added) and version V10. 

From the readme: This contains file needed to test serialization of MediaType across different versions of the code.

I had to regenerate the class for compatibility of the change of javax to jakarta so it can be loaded [here](https://github.com/mulesoft/mule-api/blob/546ac448c58bbe6cc38f2dc357922cbd50a4627c/src/test/java/org/mule/runtime/api/test/metadata/MediaTypeSerializationVersionsTestCase.java#L64.)

elrodro83: does this mean that objects serialized in previous versions of the runtime won't be able to be deserialized in the new one?

elrodro83 (changes requested):

elrodro83 (approved):

PR #1479, opened 2024-12-04
W-17050998: Deprecating MetadataService

asanguinetti:


src/main/java/org/mule/runtime/api/metadata/MetadataService.java
```diff
@@ -33,8 +33,10 @@
  * configuration.
  *
  * @since 1.0
+ * @deprecated use Mule Framework's Tooling services instead.
```
marianogonzalez, line 36: include deprecation version and more explanantion regarding how this is only kept for MTS compatibility and that will be removed when it EOLs

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #1488, opened 2024-12-10
W-17403761: Redelivery Policy - infinite loop when the redelivery is exhausted in a source configured with transactions and On Error Propagate

rbourbonnavarro:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -927,6 +927,15 @@ public final class MuleSystemProperties {
   @KillSwitch
   public static final String COMMIT_REDELIVERY_EXHAUSTED = SYSTEM_PROPERTY_PREFIX + "commit.on.redelivery.exhausted";
 
+  /**
+   * When set to true, transactions started by sources won't be committed in case of a redelivery exhausted error.
```
elrodro83, line 931: add a reference to relevant issues

elrodro83 (changes requested):

elrodro83 (approved):

PR #1533, opened 2025-02-25
W-17482887: UBP Server plugin: Spike: Integration testing

marianorean:


src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java
```diff
@@ -727,6 +727,14 @@ public final class MuleSystemProperties {
   public static final String FORK_JOIN_COMPLETE_CHILDREN_ON_TIMEOUT_PROPERTY =
       SYSTEM_PROPERTY_PREFIX + "forkJoin.completeChildContextsOnTimeout";
 
+  /**
```
IvanAndresFritzler, line 730: This is not needed, we can skip the merge of this PR and merge the other two.

IvanAndresFritzler (changes requested):

PR #1570, opened 2025-05-13
W-18463945: Supportability fuseboard - Period buffer

elrodro83:


src/main/java/org/mule/runtime/api/alert/TimedDataBuffer.java
```diff
@@ -0,0 +1,132 @@
...
+      }
+      if (next.getTime().isAfter(mark5)) {
+        agg5 = accumulator.apply(agg5, next.getData());
+      }
+      if (next.getTime().isAfter(mark1)) {
+        agg1 = accumulator.apply(agg1, next.getData());
+      }
+    }
+
+    return new TimedDataAggregation<>(agg1, agg5, agg15, agg60);
```
eze210, line 120: Does it make sense to store the now() in the aggregation?

src/main/java/org/mule/runtime/api/alert/TimedDataBuffer.java
```diff
@@ -0,0 +1,132 @@
...
+    A agg60 = baseIntevalAggregation;
+    A agg15 = baseIntevalAggregation;
+    A agg5 = baseIntevalAggregation;
+    A agg1 = baseIntevalAggregation;
+
+    final Instant mark15 = now.minus(15, MINUTES);
+    final Instant mark5 = now.minus(5, MINUTES);
+    final Instant mark1 = now.minus(1, MINUTES);
+
+    for (Iterator<TimedData<T>> it = buffer.descendingIterator(); it.hasNext();) {
```
eze210, line 105: Can this have a concurrent modification exception if you do a put during this loop?

eze210 (changes requested):

src/main/java/org/mule/runtime/api/alert/TimedDataBuffer.java
elrodro83: we´re ok with ConcurrentDeque:
```Most concurrent Collection implementations (including most Queues) also differ from the usual java.util conventions in that their [Iterators](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Iterator.html) and [Spliterators](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Spliterator.html) provide weakly consistent rather than fast-fail traversal:

they may proceed concurrently with other operations
they will never throw [ConcurrentModificationException](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/ConcurrentModificationException.html)
they are guaranteed to traverse elements as they existed upon construction exactly once, and may (but are not guaranteed to) reflect any modifications subsequent to construction.```
(https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/package-summary.html#Weakly)

eze210 (approved):

PR #1603, opened 2025-06-12
W-18438379: LockProvider API and rollout mechanism for locking

mbuchwald:


src/main/java/org/mule/runtime/api/lock/LockProvider.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.concurrent.locks.Lock;
+
+/**
+ * Provides abstraction in the creation and destruction of Mule locks.
+ *
+ * {@link LockFactory} uses instances of this interface to create and destroy locks.
+ *
+ * Lock implementation can be changed by replacing the LockProvider in the mule registry.
+ */
+public interface LockProvider {
```
eze210, line 18: Add the since

src/main/java/org/mule/runtime/api/lock/LockProvider.java
```diff
@@ -0,0 +1,28 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.api.lock;
+
+import java.util.concurrent.locks.Lock;
+
+/**
+ * Provides abstraction in the creation and destruction of Mule locks.
```
eze210, line 12: It doesn't have the destroying primitive

eze210 (changes requested):

